PR #12, opened 2015-01-14
[plugin.audio.soundcloud] 1.1.0

bromix:
I PR this one directly for Gotham (was before Frodo).

Added support for Yatse (after 2 years?). Waiting for feedback of the dev. of Yatse but I can't wait any longer :)


ronie:
tbh i don't think it's the best idea in the world that third party addons are going to distribute kodifoo libraries.
that name should be reserved for our internal python libraries imo.

bromix:
Tell me ;) Should I rename my framework :) ? Created it because of YouTube to handle the pain of possible API changes and for rapid development. Don't know how long it will take (took me 3 Month to construct the beast ;) ) and what kind of problems or errors it will cause while renaming it (Interpreter-Languages are always nice to the developer).

Let me know :)

ronie:
i'm afraid our trademark policy is pretty clear about the use of our name in third party software as well:
http://kodi.wiki/view/Official:Trademark_Policy#Related_Software

but i'll see if i can get some more opinions on this matter.
ping @tamland / @MartijnKaijser

tamland:
If you're asking me (?) I'm certainly not the person to ask about trademark policy.. But as a layman, I don't see the problem. It's framework no users will ever see. It should be completely clear kodion.. mon or anything not posted under github.com/xbmc is 'official'. I don't recall anybody having a problem or get confused by what xbmcswift or PyXBMCt is.

Other than that, consider the entire kodi* namespace reserved possible future internal use;)

ruuk:
Since this is internal to the addon's code, I don't see that there is an issue.

MartijnKaijser:
Indeed since its not visible for users I don't really have a problem with it. If would be better if it should be there to begin with or named different

PR #13, opened 2015-01-15
[plugin.video.youtube] 5.0.4

bromix:
5.0.4 (2015-01-15)
ADD: 'Live' in search
ADD: 'Play with...' for supporting external/alternative player
ADD: live events (needed some tests and feedback)
ADD: support for user names instead of channel ids
UPD: set XBMC/KODI to at least Gotham 13.2
UPD: Setup-Wizard views based on content type (KODION update)
FIX: Playlist should work again
FIX: Pagination in APIv2
FIX: settings for hiding 'Live' menu
DEL: support for Frodo


ronie:
thx!

PR #16, opened 2015-01-25
[plugin.video.netzkino_de] 2.0.5

bromix:
Direct PR to Gotham (previously Frodo)


ronie:
added, thx!

PR #20, opened 2015-02-01
[plugin.video.youtube] 5.0.5

bromix:
ADD: clear search history items
UPD: replace context menu not everywhere
FIX: Add playing video to the 'Watch History'
FIX: maybe a fix for a problem (again!) on unix systems while using python 2.7.9 + certificate verification
FIX: removed pagination for "Related Videos" - the fix before produced some nasty side-effects
FIX: Vevo signature


bromix:
The last one for today :)

ronie:
thx!

PR #33, opened 2015-03-04
[plugin.video.golem.de] 1.0.1

bromix:
Initial add


bromix:
See: https://github.com/xbmc/repo-plugins/pull/34

MartijnKaijser:
You can also force push to the same branch and it will update the PR

PR #74, opened 2015-04-14
[plugin.video.ntv-now.de] 2.2.2

bromix:
# discontinued because of DRM protection

I added only the broken tag. I don't know if it's possible to support **Adobe Access** in the future...but for now this addon won't work. Maybe remove it completely from the gotham repo? (I would be okay with me)


MartijnKaijser:
let's set it broken for now and remove in couple of months

PR #93, opened 2015-05-06
[plugin.video.rtl-now.de] 3.0.1

bromix:
Sorry for the quick update...but I have to move to helix. I noticed that some streams are only supported since Helix.


MartijnKaijser:
np

PR #111, opened 2015-06-03
Helix

kodi-bees:
plugin.video.stream-selector addon display menu to select a stream to play from several options.


ronie:
i'm wondering why this is coded as a plugin?

given the fact your addon does not provide any media and is supposed to be called by other addons,
it looks more suitable to be a xbmc.python.library imo.

PR #112, opened 2015-06-06
[plugin.video.funimation] 1.1.0

Sinap:


plugin.video.funimation/addon.xml
```diff
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8" standalone="yes"?>
-<addon id="plugin.video.funimation" name="Funimation" version="1.0.2" provider-name="Aaron Frase">
+<addon id="plugin.video.funimation2" name="Funimation" version="1.1.0" provider-name="Aaron Frase">
```
MartijnKaijser: You need to keep the same plugin id

PR #130, opened 2015-07-25
[plugin.video.syfy] 1.0.8

learningit:
Hopefully this will create just one pull request for the current commit of the syfy addon

### DIFF
--- a/plugin.video.syfy/README.txt
+++ b/plugin.video.syfy/README.txt
@@ -3,6 +3,7 @@ plugin.video.syfy
 
 Kodi Addon for Syfy
 
+Version 1.0.8 Added episode caching
 Version 1.0.7 Added show caching - should load faster
 Version 1.0.6 Added user view select, sort
 Version 1.0.5 website change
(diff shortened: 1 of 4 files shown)


learningit:
Actually how the hell do I create a separate pull request for each addon? Everytime I do a new commit it adds to the open pull request.

MartijnKaijser:
you need to create a new branch each time from the "gotham" one

learningit:
Is there a naming convention for the new branches?

MartijnKaijser:
nope, branch is what ever you call and is just a temporary name for it. It has no affect on the outcome

learningit:
Another really stupid question was it really necessary to clone the entire repo-plugins? I would think I only need the addons that I'm dealing with.

MartijnKaijser:
Yes it's only a one time clone for the entire repo. After you can just pull the latest changes and a new PR.

learningit:
Ok. Thanks for doing the commit, I was stuck because I didn't create a separate branch name for that one. I think I've got it now and can finish editting Bromix's scripts to fit my setup. Thanks!

PR #131, opened 2015-07-25
[plugin.video.foodnetwork.canada] 1.0.3

learningit:


learningit:
Martin,

I think that this commit needs to be rolled back. My pull request didn't have the folder name so it got pulled into the main repo folder. I have issued a new pull request with the folder names. I'm really not sure what happened here the folder existed in my local repo so I have no idea why the PR didn't include it in the path.

Tim

PR #132, opened 2015-07-25
[plugin.video.hgtv.canada] 1.0.2

learningit:
Hopefully this will be just for hgtv.canada - a new addon


learningit:
Yeah! It works thanks for your patience Martijn!

learningit:
Martijn,

Why is the last commit of plugin.video.foodnetwork.canada in the gotham
branch of xbmc/repo-plugins not in a folder?
I thought I buggered my fork of the repo, but I think the xbmc/repo-plugins
is where the problem is.

Thanks, Tim

PR #162, opened 2015-08-09
[plugin.video.zeemarathi] 0.0.2

virajkanwade:


MartijnKaijser:
Why have you set <language> to en ?  I doubt this plugin provides English content.

virajkanwade:
Hi Martin,

The videos are in Marathi (language of Maharashtra, India), but the listing
is in English.

I couldn't find any clear documentation, so used the standard i18n i10n en.

If thats not how thats supposed to work, will change that.

While, we are at it, is it possible to use the search from misc options to
search or scroll in my addons listing?

If you could point me to that, will submit both changes together.

Thanks
Viraj

virajkanwade:
Hi Martijn,

Could you please let me know what needs to happen for the addon to be accepted?

PR #247, opened 2015-11-20
[plugin.video.funimation] 1.2.0

Sinap:


MartijnKaijser:
@Sinap could you update the dependency versions for next plugin updates according to
http://kodi.wiki/view/Addon.xml#Dependency_versions
Till now I didn't really look at the minimum python version was specified to match the Kodi version since usually from 2.1.0 would just work fine in any Kodi version. So from now if you could set the minimum version to match the repo you are updating to that would be great.

Sinap:
@MartijnKaijser I'll be sure to do that next time I update.

PR #257, opened 2015-11-29
[plugin.video.youtube] 5.1.12

bromix:
PR for tomorrow (30.11.2015) - fixes for sort issues starting with Jarvis.


MartijnKaijser:
@bromix no need to wait till pushing updates. Since our mirror system was updated it can now handle it.

bromix:
Didn't now :) thanks for the info :)

PR #262, opened 2015-12-02
[script.module.t1mlib] 1.0.7

learningit:
This is a new library of support routines for new versions of my t1m addons. 
PRs for a few addons will follow (assuming I'm doing this right) shortly. 
I've broken some common areas of addons into the support library for ease of maintenance. 
The library requires at least Helix, but I prefer to start at Isengard, the main reason for this is subtitle handling and to ensure that most platforms have ssl support in ffmpeg (required for many of my addons).


MartijnKaijser:
thx however this one should go the the scripts repo :)
https://github.com/xbmc/repo-scripts

Since it's a module you can leave the language element empty in addon.xml

PR #267, opened 2015-12-02
[plugin.video.t1m.bravo] 2.0.3

learningit:
new Isengard version using script.module.t1mlib


MartijnKaijser:
Since you are using new addon id no user will ever get auto updated. Now we will also get duplicate Addons in repo which we don't like and users don't know which one to get. 
So how to proceed? (same for the other addons)

MartijnKaijser:
fyi keeping the same addon id is very highly preferred as that gives seamless upgrade for users

learningit:
Ok. How does that impact existing users that are not on Isengard?

MartijnKaijser:
They simply don't get these Addons

MartijnKaijser:
You could do a similar thing as this to make sure only Isengard and higher gets these newer versions https://github.com/xbmc/xbmc/blob/Isengard/addons/repository.xbmc.org/addon.xml

We use a slightly more intelligent script on our server for this

learningit:
I can go back and create new pull requests for the Isengard repo using the older addon ids. Is it alright if just change the addon id in addon.xml, or should I rename the repo folders as well? Can you delete the addons already committed to Isengard? Sorry for the confusion I thought it would be cleaner with new ids and then go back and mark the previous ones "broken".

MartijnKaijser:
Folder needs to match Addon id. If can create new PRs I'll clean up the others. 
No problem as I should have mentioned it on the first one :)

learningit:
Ok. I'll put in one to let you look at it and if it looks ok to you, I'll issue the others (probably tomorrow). thanks for your patience.

MartijnKaijser:
If you keep the old id and folder name like in gotham all should be fine. Sorry for the trouble but this makes most sense

MartijnKaijser:
Additional info. You can still mark gotham versions broken if you want as long as the Addon version is below the Isengard version

learningit:
Is it useful to kodi.org for me to push users over to Isengard? I have no problem marking the gotham versions broken (at least for the ones I have new versions for) because a number of things are corrected in the new versions. I also think that would make a bunch of the piracy boxes on earlier versions have problems, so I'm all for it.

MartijnKaijser:
I'll leave that completely up to you to decide :) 
You maintain the Addons so if you want to end gotham support that's up to you. All we ask is if it's certainly broken that we also do so in repo to prevent users trying the addon and only then finding out

learningit:
ok. Thanks Martijn.

PR #281, opened 2015-12-05
[plugin.video.wral] 2.0.1

learningit:
This is new the Isengard version replacement for plugin.video.wral. The addon id remains the same, but the name has been changed to "WCBS Programs" because the WRAL local content is not reliable any longer, but the CBS network programs are fine. Hopefully this is not an issue.


MartijnKaijser:
No problem if name changes :)

PR #321, opened 2016-01-28
[plugin.video.thinktv] 2.0.7

learningit:
version is a bit screwed in the branch name, should be plugin.video.thinktv 2.0.7. Don't think it makes a difference for the pull though. If it does let me know. Thanks


MartijnKaijser:
you can call your branch whatever you like :)

PR #328, opened 2016-02-13
[plugin.video.thenewboston] 1.1.3

lextoumbourou:
Thx guys!


MartijnKaijser:
I think you accidentally changed some file properties which shouldn't happen

MartijnKaijser:
@lextoumbourou additionally you need to bump xbmc.python to 2.14.0
Also current repo version is 1.1.2 as we set it broken. Please make it 1.1.3 or something

lextoumbourou:
Thx @MartijnKaijser. Fixed those problems and forced push. Should be ready to rock.

PR #376, opened 2016-03-07
[plugin.video.cbc] 3.0.1

learningit:


plugin.video.cbc/lib/scraper.py
```diff
@@ -0,0 +1,192 @@
...
+     name = name.decode('utf-8','replace')
+     try:
+         (name, img, infoList) = meta[sname][url] 
+     except:
+         infoList={}
+         try:
+            html = self.getRequest(url)
+            try:
+                plot = re.compile('<meta name="description" content="(.+?)"', re.DOTALL).search(html).group(1)
+            except:
```
phil65: this double-nested try-except looks bit weird :)

phil65:
also looks fine.

plugin.video.cbc/lib/scraper.py
learningit: If I recall, I just added on to fix a website change. Probably should go back and make this a bit neater in the next update. Thanks.

PR #378, opened 2016-03-07
[plugin.video.slice.canada] 3.0.1

learningit:
New video addon for Slice Canada


phil65:
might be worth redoing the icon based on fanart? quality of the icon atm looks quite bad.

phil65:
...and one thing i forgot: (applies to all plug-ins): strings.xml is the outdated way for localizing add-ons, it is now recommended to use .po files.

learningit:
With or without the redo the icon is butt-ugly.
I need to read up on the .po files, haven't used them yet. These 3 add-ons(slice, showcase, etcanada) would be good for that as Canadians do things bilingually.

phil65:
There  is some tool available somewhere which converts  the xml files to po files.... Would have to searcch for it though.

learningit:
Don't bother. I want to understand how they work. These add-ons have very limited use of language strings anyway.

MartijnKaijser:
How they work is basically the same as the xml but just in different format. http://kodi.wiki/view/Language_support#Layout_of_the_strings.po_file

In Isengard we also changed folder layout for language with backwards compatibility. 
We now use the ISO code in the folder names. https://github.com/xbmc/xbmc/tree/master/addons/skin.estuary/language as example

PR #380, opened 2016-03-07
[plugin.video.iplayerwww] 2.1.3

CaptainTK:


plugin.video.iplayerwww/resources/lib/ipwww_video.py
```diff
@@ -813,50 +785,23 @@ def AddAvailableLiveStreamsDirectory(name, channelname, iconimage):
...
             color = 'green'
-        elif bitrate > 1000:
+        elif bitrate > 1.0:
             color = 'yellow'
-        elif bitrate > 600:
+        elif bitrate > 0.6:
             color = 'orange'
         else:
             color = 'red'
-        # title = name + ' - [I][COLOR %s]%0.1f Mbps[/COLOR] [COLOR white]%s[/COLOR] [COLOR grey]%s[/COLOR][/I]' % (color, bitrate/1000.0, supplier, device)
```
phil65: I would recommend to not use color names since those are skin dependent... better take the real AARRGGBB values.

phil65:
Rest looks fine.

CaptainTK:
Thanks for the hint regarding colours. I'll change this for the next release.

Would you mind merging it the way it is once again? The colour values exist like this since the first version of the add-on.

phil65:
Sure sure.

PR #384, opened 2016-03-13
[plugin.video.animeftw] - removed because it is against our non-pirac…

Memphiz:
…y-stance.

As it was stated by Ned here:

http://forum.kodi.tv/showthread.php?tid=263000

This addon is violating our no-piracy-stance and shouldn't be available through our official repo.

This PR removes that addon from our repo.


ronie:
@Memphiz that forum link does not work?

we've received a email from FTW Entertainment LLC about why their addon was removed.
according to this forum post, they only distribute public domain content:
http://forum.kodi.tv/showthread.php?tid=255179

Memphiz:
i realised that the forum link is pointing to nowhere (no clue why - this forum thread was there!) - i also contacted @NedScott already -and yes i got a couple of mails too...

ScudLee:
Not licensed != Public domain

In this context "not licensed" just means that no (typically) US company has paid for the rights to distribute those shows in English.  It does not mean animeftw has the rights to distribute them.  They don't, and they even admit as much themselves:
> All anime listed in FTW sites are just for promotion and to encourage anime fans to buy originals when it release. We here don't have any affiliate with the anime creator or licenser. [sic]

https://www.ftwentertainment.com/legal

A legitimate streaming site would brag about the licenses they _have_, not that they only have unlicensed shows.  Fansubs are still a form of piracy, even if the shows are unlicensed.

...Even then, most of the material listed on the site is still blatantly licensed by other companies!
The claim to having taken "significant steps" toward removing licensed content is dubious at best, and is probably just a reference to the DMCA compliance page, rather than any direct action on their part (but either way it's _still_ purely piracy.)

ronie:
> Not licensed != Public domain

i stand corrected :-)

'unlicensed' is the term used in the forum post.

PR #386, opened 2016-03-19
[plugin.video.soap4.me] 0.3.1

ufian:
New video addon for site http://soap4.me


ufian:
I think I did something strange. Should I create pull-request again without commits from master?

MartijnKaijser:
Sorry but seems like a complete illegal streaming site.

PR #390, opened 2016-03-22
[plugin.video.tviplayer] 0.1.0

darksky83:
TVI Player addon for kodi

Independent and open source plugin not endorsed by TVI


darksky83:
Commit without .gitignore file was requested

learningit:
In future, please use strings.po, rather than strings.xml. .Info on .po file is here: http://kodi.wiki/view/Language_support#What_is_strings.po

PR #392, opened 2016-03-25
[plugin.video.tviplayer] 0.2.0

darksky83:
* Bug fix for direrent flash player use in some Shows
* Change from strings.xml to strings.po


Razzeee:
Should be fine.

plugin.video.tviplayer/resources/ondemand.py
```diff
@@ -246,6 +240,7 @@ def tvi_resolver(url):
     try:
         source = abrir_url(url)
     except:
+        print "Unexpected error:", sys.exc_info()[0], sys.exc_info()[1]
```
learningit: Please remove print statement, use log() instead

plugin.video.tviplayer/resources/ondemand.py
```diff
@@ -261,6 +256,7 @@ def tvi_resolver(url):
             if match[0][0]:
                 return match[0]
         except:
+            print "Unexpected error:", sys.exc_info()[0], sys.exc_info()[1]
```
learningit: Please remove print statement, use log() instead

darksky83:
I will create a new PR whit the corrections.

PR #401, opened 2016-03-27
[plugin.video.kermodeandmayo] 0.5.2

LS80:


learningit:
Looks good to me. Only comment is that in the next update, please use strings.po instead of strings.xml.

PR #407, opened 2016-03-28
[plugin.video.on_aol] 1.1.1

divingmule:
Fix an encoding mix-up and add strings.po file.

@learningit  This looks better, I think.

Thanks.


learningit:
Yep much better, thanks.

PR #409, opened 2016-03-29
[plugin.video.digistorageplayer] 1.0.0

marianpavel:


phil65:
Please use x.y.z version numbers.
Why did you add the complete requests module? You could simply use the module which is available on official repository.

marianpavel:
This is the way I made him to work with requests :(
I had some errors when I tried to use it with requests from kodi but I think I haven't done the import as I should.

phil65:
Example:
https://github.com/phil65/repo-plugins/blob/updates_helix/plugin.video.nfl.gamepass/addon.xml#L13
https://github.com/phil65/repo-plugins/blob/updates_helix/plugin.video.nfl.gamepass/resources/lib/pigskin.py#L15

Make sure to have that module add-on installed on your local machine (it will auto-install when installing from repository)

That´s all you need.

marianpavel:
I have made the changes and tested, should I make another pull request or we are going with this one ?
Thank you

plugin.video.digistorageplayer/addon.py
```diff
@@ -0,0 +1,104 @@
...
+from xbmc import log as xbmc_log
+from urlparse import parse_qsl
+
+requests.packages.urllib3.disable_warnings()
+
+base_url = sys.argv[0]
+addon_handle = int(sys.argv[1])
+args = urlparse.parse_qs(sys.argv[2][1:])
+
+xbmcplugin.setContent(addon_handle, 'movies')
```
phil65: Are you sure that "movies" is a good decision for content type? I would suggest to use "files" i think.

phil65:
no need for another pull request. thx for removing the library.

marianpavel:
I have changed the content type to files and tested it, thank you for the suggestion :+1:

plugin.video.digistorageplayer/addon.xml
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<addon id="plugin.video.digistorageplayer" name="Digi Storage Player" version="1.0.0" provider-name="RCSRDS">
+	<requires>
+		<import addon="xbmc.python" version="2.14.0"/>
+		<import addon="script.module.requests" version="2.9.1"/>
+	</requires>
+	<extension point="xbmc.python.pluginsource" library="addon.py">
+		<provides>executable</provides>
```
phil65: do you want it listed in programs section? otherwise you could change it to "video".

phil65:
One last comment, apart from that everything looks good to me. Thx!
@learningit feel free to have another look and pull if done.

plugin.video.digistorageplayer/addon.xml
marianpavel: Yes, I want it to programs section.

phil65: ok.

plugin.video.digistorageplayer/addon.xml
```diff
@@ -0,0 +1,18 @@
...
+	</extension>
+	<extension point="xbmc.addon.metadata">
+		<platform>all</platform>
+		<summary lang="en">Digi Storage Player</summary>
+		<description lang="en">Digi Storage Player</description>
+		<license>GNU General Public License, v2</license>
+		<language></language>
```
learningit: Please add <source> entry

plugin.video.digistorageplayer/addon.xml
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<addon id="plugin.video.digistorageplayer" name="Digi Storage Player" version="1.0.0" provider-name="RCSRDS">
+	<requires>
+		<import addon="xbmc.python" version="2.14.0"/>
```
learningit: If you want this available in Jarvis and above versions, "xbmc.python" should be set to "2.24.0" , your current setting is for Gotham. Please see: http://kodi.wiki/view/Addon.xml#Dependency_versions for more info.

learningit:
Ok. I'm pretty confused now. You need to squash all your commits down into one, so that I can finish looking at this. It's pretty hard to do as you're doing commits on the fly while I'm trying to read the code.
Please see: https://github.com/ginatrapani/todo.txt-android/wiki/Squash-All-Commits-Related-to-a-Single-Issue-into-a-Single-Commit
for how to squash them down, looks like you have 6 to squash at this point.

plugin.video.digistorageplayer/resources/settings.xml
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="utf-8" standalone="yes"?>
+<settings>
+    <setting id="username" type="text" label="Username" value="" />
```
learningit: Normally we require that labels used in settings.xml be translatable strings in a strings.po file. Because the labels you're currently using are self-explanatory you can leave as is. Please keep this in mind for the future.
Please see: http://kodi.wiki/view/Language_support#What_is_strings.po for more info.

marianpavel:
I don't quite understand how to do it, I am getting an error when trying to squash them into single commit.

Razzeee:
Are you using commandline for a client for git?

marianpavel:
For commits and branches I use the desktop application

marianpavel:
Ok I have solved it @Razzeee , thank you

learningit:
Not quite solved, you didn't include the original commit in the squash.

marianpavel:
I made another commit and 6 became 7 :+1: now there is only 1 commit, thanks.

learningit:
Good to go.

marianpavel:
Thank you, one question I have, when it will be available in kodi ?

phil65:
might take a day perhaps.

Razzeee:
Really depends on which time zone your in ;)
Generally speaking it's updated each moring, not sure in which the zone the server is that updates it.

PR #411, opened 2016-03-31
[plugin.video.wnbc] 3.0.1

learningit:


plugin.video.wnbc/lib/scraper.py
```diff
@@ -0,0 +1,110 @@
...
+        if episode: infoList['Episode'] = int(episode)
+        duration = b.get('totalDuration')
+        if duration: infoList['Duration']= int(duration/1000)
+        airDate = int(b['firstAiredDate'])
+        infoList['date'] = datetime.datetime.fromtimestamp(airDate).strftime('%d.%m.%Y')
+        airDate = datetime.datetime.fromtimestamp(airDate).strftime('%Y-%m-%d')
+        infoList['aired'] = airDate
+        infoList['premiered']  = airDate
+        infoList['year'] = int(airDate.split('-',1)[0])
+        infoList['MPAA'] = b.get('rating')
```
phil65: also add infoList['mediatype'] = "episode"

plugin.video.wnbc/resources/settings.xml
```diff
@@ -0,0 +1,21 @@
...
+<settings>
+
+	<category label="30012">
+		<setting type="sep" />
+		<setting id='enable_meta' type ="bool" label="30013" default="true"/>
+           <setting id="init_meta" type="bool" label="30014" default="true" enable="!eq(-1,false)"/>
+		<setting type="sep" />
+	</category>
+
+  <category label="30020">
```
phil65: not a big fan of this hack..

learningit: I prefer the term "workaround". As discussed this is necessary in Isengard because the core doesn't set the view type for all levels in the content type.

phil65: No, it´s a quite dirty hack. You could argue that this is nice to have, but it definitely is not "neccessary".

learningit: I can remove it, but there will be many user complaints. Shall I direct them to you?

phil65: Sure, I love dealing with them :)
Anyways, although I think that the more experienced plugin authors should lead by example, this is no blocker, no need to close the PR. I wont mention again that I dislike this hack in any of your future PR´s . ;)

Rest from code looks fine btw.

PR #421, opened 2016-04-02
[plugin.video.twitch] 1.4.0

MrSprigster:


plugin.video.twitch/changelog.txt
```diff
@@ -88,3 +88,9 @@
 - added ability to play new type twitch VODs + major code refactoring/optimisation and bug fixes (now utilising python-twitch [0.3.8] by ingwinlu). - ingwinlu
 1.3.1
 - fixed bug: addon would not play non-transcoded streams if requested quality was lower than source
+1.3.2
+- added followed games Thanks to stevensmedia
+1.3.3
+- added French language
+1.4.0
```
Razzeee: One minor thing:
The changelog should be in the opposite order. So that the last version is always on top, getting older the more you get to the bottom.

plugin.video.twitch/resources/language/Danish/strings.xml
```diff
@@ -0,0 +1,61 @@
+﻿<?xml version="1.0" encoding="utf-8" standalone="yes"?>
```
Razzeee: You should move these strings.xml to strings.po
Not needed for inclusion, but the xml are deprecated and will stop working sooner or later

plugin.video.twitch/addon.xml
```diff
@@ -1,27 +1,30 @@
...
-<addon id='plugin.video.twitch' version='1.3.1' name='TwitchTV' provider-name='StateOfTheArt and ccaspers'>
-  <requires>
-    <import addon='xbmc.python' version='2.1.0'/>
-    <import addon='script.module.simplejson' version='2.0.10'/>
-    <import addon="script.module.xbmcswift2" version="1.2.0"/>
-  </requires>
-  <extension point='xbmc.python.pluginsource' library='default.py'>
+<addon id='plugin.video.twitch' version='1.4.0' name='Twitch' provider-name='StateOfTheArt and ccaspers'>
+    <requires>
+        <import addon='xbmc.python' version='2.20.0'/>
```
learningit: You are asking for this PR to be pulled into the "Gotham" repo, but you have the 'xbmc.python' version set to Isengard. Which version repo is this update intended for?

MrSprigster: It is intended to support versions as far back as Gotham.

learningit: If you intend to support Gotham, you need to set 'xbmc.python' to '2.14.0', the current setting will fail for versions below Isengard.

MartijnKaijser:
already pushed

PR #429, opened 2016-04-10
[plugin.video.la3] 1.0.0

NeverWise:
Hi,

Sending new add-on.
Let me know if there are any issues.

Thanks!


plugin.video.la3/addon.xml
```diff
@@ -0,0 +1,25 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<addon id="plugin.video.la3" name="La3" version="1.0.0" provider-name="NeverWise">
+  <requires>
+    <import addon="xbmc.python" version="2.14.0"/>
```
phil65: jarvis is 2.24.0

phil65:
I am not sure if this qualifies as a plugin...

NeverWise:
This plugin works also in gotham. If you want I can push it in the gotham repository but in readme.md file of gotham you can read:
 > New add-on additions: Denied

I have pushed other plugin like this (plugin.video.skytg24 and plugin.audio.resetradio).

phil65:
Send it to the "lowest" active one. Probably helix atm? @learningit ?

learningit:
For new addons, I believe Helix is the current lowest as per @phil65 above.  You need to set 'xbmc.python' to the correct value for the lowest version as well. See http://kodi.wiki/view/Addon.xml#Dependency_versions for details.

learningit:
Hmm after looking at the code in default.py, I would tend to agree with @phil65 that this is probably better called a script than a plugin. I'm basing that on the fact that it never creates any directory and will confuse the hell out of anything that uses json to parse for plugin functions.

phil65:
@tamland opinion would interest me.

ronie:
a video plugin provides videos through a vfs. anything else is a script.

NeverWise:
As ronnie says, my plugin provides an italian tv stream (trying is believing).
As I already said, my plugin works on gotham and in the page linked by @learningit gotham is not marked as deprecated.
So I think the plugin is right. Let me know in which version I have to push it.
Last thing, if you read this [link](https://github.com/xbmc/repo-plugins/blob/helix/README.md) you can find
> New add-on additions: Denied

phil65:
You do not even need the xbmcplugin module, so imo this doesnt qualify as a plugin.
Please PR this to repo-scripts repository. Agreed, @ronie ?

MartijnKaijser:
addon id has plugin in it so don't make it more difficult than it already is and keep it in plugin repo

learningit:
This is the current table of for allowed add-on submissions. http://kodi.wiki/view/Submitting_Add-ons#Allowed_submissions . So assuming that we're going to put this in as a plugin, the addon.xml and the PR still need to be set to a minimum of Isengard by the author in order for us to process teh pull request.

ronie:
> Agreed, @ronie ?

yup this is not, i repeat, not a plugin.
having it defined as a plugin in addon.xml will only to errors in the logfile.

NeverWise:
Hi, sorry I'm late.
I have made some changes to the code to solve the error quoted by @ronie and I have updated the xbmc.plugin version to isengard.
Is it necessary a new PR or it's enough a commit?

Razzeee:
It has to be a new PR, as we need you to send it to one of the other repos, as this is not a plugin :) 
Without looking into it, probably the script repo

NeverWise:
I have done some changes.
Can be accepted as a plugin now? If no, what's wrong?

phil65:
This is no plug-in. A plug-in is browsable, there should at least be a root folder with listitems.

NeverWise:
Ok, for now that's all.

Thank you.

PR #433, opened 2016-04-14
[plugin.video.iplayerwww] 2.2.0

CaptainTK:


learningit:
Two general points to please consider fixing before your next update:
1) Please avoid the use of \__name\__ style labels. These are meant for python internals.
2) Please make use of strings.po rather than strings.xml.  Strings.xml is deprecated.

PR #436, opened 2016-04-18
[plugin.video.comettv] 3.0.4

learningit:


Razzeee:
Looks good

plugin.video.comettv/resources/lib/scraper.py
```diff
@@ -0,0 +1,58 @@
...
+         et = datetime.datetime.fromtimestamp(float(b['EndTime'])).strftime('%H:%M')
+         infoList['duration'] = int(float(b['EndTime']) - float(b['StartTime']))
+         name = b['Title']
+         if b['EpisodeTitle'] != '':
+            infoList['TVShowTitle'] = name
+            try: infoList['Season'] = int(b['TVObject']['SeasonNumber'])
+            except: pass
+            try: infoList['Episode'] = int(b['TVObject']['EpisodeNumber'])
+            except: pass
+         plot = '%s - %s                 %s min.\n[COLOR blue]%s[/COLOR]\n%s' % (st, et, infoList['duration'] / 60,  b['EpisodeTitle'], b['CopyText'])
```
phil65: \n should probably be [CR]?

learningit: I have no idea. Where are all these funny [..] sequences defined?

phil65: http://kodi.wiki/view/Label_Formatting

learningit: Well, in the less enlightened age that I come from: a CR returns to the start of a line, A LF scrolls down a line and a \n newline does both or is interpreted as necessary by the underlying terminal device driver.  At least it was that way for a number of commercial c compilers I wrote.

More seriously, does label processing not manage control chars from Python? if not where are things like \t ?

phil65: I think for kodi infolabels we should never "scroll down a line". Such kind of formatting wont work properly with lot of skins (--> varying textbox widths)

learningit: I suppose add-on authors have to live within the constraints that skinners create. I'll fix this when I get to a PC later tonight. I probably should get rid of those try / except clauses and add mediatype as well.

phil65: Cool. Thx.

PR #444, opened 2016-04-23
[plugin.video.thinktv] 3.0.3

learningit:


plugin.video.thinktv/resources/lib/scraper.py
```diff
@@ -0,0 +1,277 @@
...
+        ilist = self.addMenuItem(name,'GC', ilist, url, thumb, fanart, infoList, isFolder=True, cm=contextMenu)
+    if pgNum != '': ilist = self.addMenuItem('[COLOR blue]%s[/COLOR]' % addonLanguage(30050),'GS', ilist, genreUrl+'|'+str(int(pgNum)+1), self.addonIcon, self.addonFanart, None, isFolder=True)
+    return(ilist)
+
+
+ def getAddonCats(self,url,ilist):
+    addonLanguage  = self.addon.getLocalizedString
+    self.doPBSLogin()
+    thumb  = xbmc.getInfoLabel('ListItem.Art(thumb)')
+    fanart = xbmc.getInfoLabel('ListItem.Art(fanart)')
```
phil65: Using getInfoLabel() always feels quite hack-ish.. what do you need these 2 for?

phil65:
One comment, apart from that all looks fine.

plugin.video.thinktv/resources/lib/scraper.py
learningit: I tend to agree about that, but the I wasn't really happy with the previous way I was doing this either. The issue here is that you have a level of directory which contains tv shows, which goes to a level to select episodes, clips or previews, which then goes to a level to select individual episodes or videos. Previously I rolled the thumb, fanart and tvshow name into the URL for the plugin. They get pretty long and ugly and occasionally have some unicode issuses which need to be resolved because urllib.quote or quote_plus use strings and don't like unicode. So here I tried using the getInfoLabel calls to retreive them from the selected TV show item. There are issues obviously when the URL is used from favourites or backing through the directories.  The same issues exist when the images are not static (believe it or not, some image requests have a timestamp & autho field which expires, though caching usually deals with that). I thought about using window properties as well to hold the values, but the favourite issue still remains. Not sure if I'm describing this issue in an understandable way.

phil65: yup, understood the problem.
perhaps unicode and similar problems could be avoided by base64-encoding/decoding the information? (just an idea, feel free to merge this without any adjustments)

learningit: the base64 idea is very good. I'm going to merge this for the time being and longer term work some thing generic out using base64 back in the helper routines library as it is a common problem with my add-ons. Thanks

PR #454, opened 2016-05-03
[plugin.video.putiov2] remove broken plugin

igungor:
This plugin is broken and is not maintained anymore. We have plugin.video.putio plugin already and it's maintained pretty well.


Razzeee:
Removal is a no go. If anything, it should be set as broken via the addon.xml

igungor:
Oh, didn't know that, thanks. I'll open a new PR and mark this plugin as broken.

PR #458, opened 2016-05-05
[plugin.video.putio] 2.1.0

igungor:
* Sort files by date
* Inform user if payment is expired
* Don't show delete context for shared files
* Fix errors dialogs popping up occasionally while watching a video
* Fix AttributeError where an audio file has no 'start_from' attribute


plugin.video.putio/main.py
```diff
@@ -60,34 +60,47 @@ def populate_dir(files):
...
 
-        # url when a delete action is triggered
-        delete_ctx_url = build_url(action='delete', item=item.id)
+        # http://kodi.wiki/view/InfoLabels
+        type_ = 'video' if item.is_video or item.is_folder else 'music'
+        li.setInfo(type=type_,
+                   infoLabels={
+                       'date': item.created_at.strftime('%d.%m.%Y'),
+                       'size': item.size,
+                       'title': item.name,
```
phil65: if this is used for videos, please also set "mediatype".

phil65:
one comment, rest looks good. Thx!

plugin.video.putio/main.py
igungor: done.

PR #462, opened 2016-05-05
[plugin.video.antennasicilia] 1.0.0

RosarioACali:


Razzeee:
Please read the rules, you will need to change your commit message. And I don't think we can accept submodules, we will need you to push the actual code.

learningit:
I'm closing this as we don't accept submodule commits.

phil65:
@RosarioACali the problem probably was that you also committed your .git folder. If you dont add that folder, all should be fine.

PR #473, opened 2016-05-16
[plugin.video.nba] 0.8.1

maxgalbu:


plugin.video.nba/common.py
```diff
@@ -104,6 +104,8 @@ def login():
         littleErrorPopup( xbmcaddon.Addon().getLocalizedString(50022) )
         return ''
 
+    log("Login reponse: %s" % content, xbmc.LOGNOTICE)
```
phil65: LOGDEBUG, please.

plugin.video.nba/nbatvlive.py
```diff
@@ -62,70 +107,129 @@ def playLiveTV():
...
+            try:
+                request = urllib2.Request(video_play_url, None, {'Cookie': vars.cookies})
+                response = urllib2.urlopen(request)
+                content = response.read()
+            except urllib2.HTTPError as e:
+                log("nba live tv: failed getting url: %s %s" % (video_play_url, e.read()))
+                littleErrorPopup( xbmcaddon.Addon().getLocalizedString(50023) )
+                return
+
+            if not content:
```
learningit: I believe that here you are checking whether a variable (content) exists or not.  This is not a correct way to do that. The correct way is:

if not 'content' in vars() \:
or
if not  'content' in globals()\:

plugin.video.nba/nbatvlive.py
```diff
@@ -62,70 +107,129 @@ def playLiveTV():
...
-        path = urllib.quote_plus(str(match.group(2)))
-        querystring = match.group(3)
-        video_play_url = "http://%s/play?url=%s&%s" % (domain, path, querystring)
-        log("nba live tv: play url is %s" % video_play_url, xbmc.LOGDEBUG)
+        item = xbmcgui.ListItem(path=video_url)
+        xbmcplugin.setResolvedUrl(int(sys.argv[1]), True, item)
+
+    @staticmethod
+    def playLive():
+        if not vars.cookies:
```
learningit: Another example of testing for var existence.

plugin.video.nba/nbatvlive.py
maxgalbu: I'm checking whether a variable is an empty string or not, not whether it exists or not. In this case the http request can return an empty string, but the variable should be defined

plugin.video.nba/nbatvlive.py
maxgalbu: This is the same case as before, the variable is defined (here: https://github.com/maxgalbu/xbmc.plugin.video.nba/blob/master/vars.py#L32) and I'm just checking whether it's empty or not.

learningit: Sorry, I looked for the def and didn't see it. Looked like you were checking for fallout from the try preceding the check. Rest looks ok.

PR #482, opened 2016-05-19
[plugin.video.twit] 2.1.5

divingmule:
This update is a workaround for a SSL bug in Python < v3.2, thus the hard coded shows.py file. Luckily that info doesn't change very often.

Thanks!


learningit:
The current version of twit.tv is in the Jarvis repo, but you want to commit this to Helix?

divingmule:
Humm... I'm not seeing it in Jarvis.  I'm finding the last version in Gotham. Some info label stuff changed with Helix that is why I wanted to move it there.

learningit:
I saw it wasn't in Helix and checked here http://addons.kodi.tv/show/plugin.video.twit/ which has 2.1.3 in Jarvis.

learningit:
Nevermind, it's in both gotham and jarvis. Ok so I'll update into Helix.

divingmule:
Thanks!

learningit:
Damn, I was too quick to merge. In addon.xml xbmc.python needs to be set to 2.19.0 for Helix. Please fix this as it will mess up folks who direct download and install on Gotham.

PR #494, opened 2016-05-28
[plugin.video.worldstarcandy] 1.0.0

skipmodea1:


phil65:
I am not sure if we allow tits-and-asses content. @ronie, do we have a guideline for stuff like this?

ronie:
nope, nothing written in stone afaik.
i don't recall anyone ever submitted an adult content addon before.

i'd prefer to keep our addon repo suitable for all ages ;-)

skipmodea1:
Just to be clear here, the kodi-team does not allow an addon with adult content in the kodi repo?

phil65:
Correct.

skipmodea1:
Right. Closing the PR.

PR #500, opened 2016-05-31
[plugin.video.roosterteeth] 1.3.1

skipmodea1:


plugin.video.roosterteeth/.gitattributes
```diff
@@ -0,0 +1,22 @@
+# Auto detect text files and perform LF normalization
```
learningit: Please remove the gitattributes and gitignore files from the commit.

skipmodea1: removed the gitattributes and gitignore files.

skipmodea1:
I removed the gitattributes and gitignore files.

learningit:
I missed this before: I think you should be using xbmc.LOGDEBUG rather than xbmc.LOGNOTICE for your debug messages. The rest looks OK to me.

skipmodea1:
Changed debug stuff

PR #504, opened 2016-06-03
[plugin.video.rivedila7] 0.0.6

luivit:
hi, bugfix
thx


plugin.video.rivedila7/addon.py
```diff
@@ -57,18 +57,21 @@ def rivedi_la7():
             else:
```
learningit: Why not use for div in giorno[1:] in the above, rather than the use of i?

luivit:
updated, it was a temporary solution that then I forgot to adjust

PR #505, opened 2016-06-03
[plugin.video.hgtv] 3.0.6

learningit:


phil65:
same here, solid bg for fanart please, rest looks good. :)

learningit:
Ok, fixed the fanart background to solid.

PR #515, opened 2016-06-06
[plugin.video.blitzr.music] 1.0.0

lambda81:
video addon for
http://blitzr.com


plugin.video.blitzr.music/addon.xml
```diff
@@ -0,0 +1,22 @@
+﻿<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<addon id="plugin.video.blitzr.music" version="1.0.0" name="Blitzr Music" provider-name="lambda">
+	<requires>
+		<import addon="xbmc.python" version="2.24.0" />
+		<import addon="plugin.video.youtube" optional="true" />
+	</requires>
+	<extension point="xbmc.python.pluginsource" library="addon.py">
+		<provides>video audio</provides>
```
phil65: there should only be one section, not both video and audio.

phil65:
solid background for the fanart image, please.
Since you seem to re-use lot of stuff across all your different add-ons, it would be nice to move that kind of stuff into a separate helper module so we dont have to review the same code over and over again.

lambda81:
i will send a new pull request for this addon.
with a new fanart and as a video addon only.

you are totally right about the need of a module.
i wasn't expecting to end up with this number of addons.
it will be easier for me too to maintain them.
what do you consider as the best option?
should i close the pull requests of all 15 addons, create a module and send 15 new pull requests with it or should i create a module and start using it on the future updates of each addon?

phil65:
I think the best way is to close your PRs, create a helper module (let's say script.module.lambdahelpers) and then create PRs for the helper module and some very few first add-ons. If they are fine then you can create PRs for the rest.
Thx for cooperating. ;)

lambda81:
i will create it and send 2-3 each time.
thanks for your help too.

PR #516, opened 2016-06-06
[plugin.video.antenna.gr] 1.0.0

lambda81:
video addon for
http://www.antenna.gr

ANT1 is a FTA Greek network
https://en.wikipedia.org/wiki/ANT1


phil65:
Looks good in general.
Only thing I dont like is the quite excessive usage of try-except-pass. You will probably hide some bugs that way.

lambda81:
I can make the needed changes if required.
Even if the addon is accepted with its current code,
I will use it less on next updates of the addon.

PR #523, opened 2016-06-09
[plugin.audio.eradio.gr] 1.0.0

lambda81:
radio addon for
http://www.e-radio.gr


phil65:
that's much easier to review now. :)
Looks good to me.

lambda81:
phil can i ask you something here?
it's not about this addon, but it's better asking here before making a pull request
it's about service.subtitles.subtitles.gr
https://github.com/xbmc/repo-scripts/pull/186
it's not a new subtitle service
i have it coded about 2 years ago
i want to send an updated version
can i use the module i created on it?
what i mean
the first versions of the service had been added to gotham
the module is new, so it's jarvis or higher
what are my options?
can i send version 1.2.0 to jarvis?
if i send it to gotham i won't be able to use the module

phil65:
yeah you can send the updated version to jarvis branch and use the module then. Older kodi versions will still get the previous subtitle addon then.

lambda81:
thanks for your reply, i will send it to jarvis

PR #527, opened 2016-06-09
[plugin.video.alphatv.gr] 1.0.0

lambda81:
video addon for
http://www.alphatv.gr

ALPHA TV is a FTA Greek network
https://en.wikipedia.org/wiki/Alpha_TV


plugin.video.alphatv.gr/addon.xml
```diff
@@ -0,0 +1,25 @@
+﻿<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<addon id="plugin.video.alphatv.gr" version="1.0.0" name="ALPHA Player" provider-name="lambda">
+	<requires>
+		<import addon="xbmc.python" version="2.24.0" />
+		<import addon="script.module.lamlib" version="1.0.0" />
+		<import addon="plugin.video.youtube" optional="true" />
```
phil65: @learningit do we support an "optional" attribute? Never seen that before.

learningit: The old VEVO add-on had the same "optional" for YouTube. No idea if it's valid or not, doesn't seem to hurt anything.

lambda81:
hi phil, i've checked the wiki
http://kodi.wiki/view/addon.xml#optional_attribute

lambda81:
if it's something not used anymore or deprecated, i could send new pull requests and remove the optional attribute.

phil65:
seems to be fine.

plugin.video.alphatv.gr/addon.xml
MartijnKaijser: optional is working iirc. Might as well remove it

learningit: Conversation about "optional" aside, I believe that you should specify a version when using it.

learningit:
The "optional" attribute is ok. You do need to set the version for it though - also in your other PR [plugin.video.ert.gr] 1.0.0

lambda81:
thanks learningit for your help.
should i close these PRs and send new ones?
what is your recommendation for the version i should use?
5.2.1?

learningit:
I actually don't use the YT add-on in the repo, so I'm not sure what the latest version is. You can look here to find it: http://addons.kodi.tv/

You do not need to close the PRs. Just make your changes and commit them to the your repo.  You then need to squash them in to a single commit (per add-on) and force an update. Instructions for how to squash and update them are here: 
https://github.com/ginatrapani/todo.txt-android/wiki/Squash-All-Commits-Related-to-a-Single-Issue-into-a-Single-Commit

lambda81:
thanks learningit, i checked it's 5.2.1
i have already deleted the previous repos before reading your reply.
can i make new ones?

learningit:
Yes, of course.

PR #534, opened 2016-06-11
[plugin.video.blitzr.music] 1.0.0

lambda81:
video addon for
http://blitzr.com


plugin.video.blitzr.music/addon.py
```diff
@@ -0,0 +1,116 @@
...
+
+import urlparse,sys
+
+from resources.lib import blitzr
+
+params = dict(urlparse.parse_qsl(sys.argv[2].replace('?','')))
+
+
+try:
+    action = params['action']
```
phil65: small tip: use action = params.get("action"), then you dont need the try-except.

phil65:
looks good to me.

plugin.video.blitzr.music/addon.py
lambda81: thanks phil
you are totally right, i have to use it as less as possible
i have removed try except from the main fuctions on my latest PRs
eventually i will keep in only on some for loops

PR #551, opened 2016-06-15
[plugin.video.trakt.trailers] 1.0.0

lambda81:
trailer addon for
https://trakt.tv/


phil65:
Trakt stuff is already part of script.extendedinfo, not sure if we want to duplicate that.

lambda81:
extendedinfo is "the "Swiss Knife" of info:)
If you need any kind of info, trailers, youtube videos, everything, it's there.
I am using extendedinfo all the time, i know it has access to trakt database and i was thinking the same.

The logic behind this addon is to have a "simple" trakt only addon for traliers that can also send a movie or tv show to trakt account from context menu.
Like other databases have (imdb trailers, apple trailers etc)

phil65:
I think to get that in it should at least surpass extendedinfo trakt part in terms of capabilities. Did you implement all functionality which is part of extendedinfo? I saw that you implemented some log-in stuff, that might be an argument. As an idea, a context menu entry for "similar movies" would also be a nice addition I think.

@learningit , what's your opinion?

lambda81:
i haven't implement most of the functionality which is part of extendedinfo, there is no reason for that, extended info has every kind of info i would ever need.
this is meant as a simple addon, not something extended like extended info
one click trailer playback from trakt, the same way other trailer addons work (imdb traiilers etc)
if user wants something more complete, he should use extended info

lambda81:
i have been missing extended info option to play the trailer instead of showing the dialog!
i was trying to see what phil is talking about and checked extended info settngs
i am a heavy extended info user, i should have noticed it:)
with this option disabled, extended info works the same way.
i wish i would have noticed the option before working on it.
phil it would make sense if this addon won't be merged.

PR #552, opened 2016-06-16
[plugin.video.thinktv] 3.0.5

learningit:


plugin.video.thinktv/resources/lib/scraper.py
```diff
@@ -16,91 +16,97 @@
...
-   except: cj=[]
-   for cookie in cj:
-       if cookie.name == 'sessionid':
-          badCookie = cookie.is_expired()
+    if self.addon.getSetting('enable_login') != 'true':
+        return
 
-   if badCookie:
+    profile = self.addon.getAddonInfo('profile').decode(UTF8)
+    pdir  = xbmc.translatePath(os.path.join(profile))
```
phil65: nitpick: os.path.join with a single argument doesnt make too much sense :)

plugin.video.thinktv/resources/lib/scraper.py
```diff
@@ -16,91 +16,97 @@
...
+    if not os.path.isdir(pdir):
+        os.makedirs(pdir)
+    cjFile = xbmc.translatePath(os.path.join(profile, 'PBSCookies.dat'))
```
phil65: I think this should have a .decode("utf-8") in order to deal with special chars correctly.

phil65:
rest looks fine :)

plugin.video.thinktv/resources/lib/scraper.py
learningit: yeah, bad C&P

plugin.video.thinktv/resources/lib/scraper.py
learningit: Actually never mind the above, after looking at the code it's a bit retarded to be saving the cookie  jar instead of just saving the sessionid cookie as a setting. I'll rewrite. thanks.

learningit: Ooops, no it wasn't so retarded, I need to maintain cookie state across multiple calls. So I guess I need to ask the question again:
Do you mean before or after the xbmc.translatePath()? I thought the decode on profile in line 28 was enough, or does the output of xbmc.translate need the decode as well?

learningit:
I think that should address the unicode filename issues.

PR #574, opened 2016-06-28
[plugin.video.espn_3] 2.9.8

locomot1f:


learningit:
You should create a separate working repo, do all your changes from various contributors, squash the commits and then do a PR to update the xbmc/repo-plugins. Not cool to use xbmc/repo-plugins as your working copy.

locomot1f:
we do apologize. we don't want to cause problems.
we're working on a procedure to update the addon. it has never been our intention to use xbmc as our testing platform. 
case of too many chefs in the kitchen. we're getting it together. 
this won't happen again.

PR #578, opened 2016-07-02
[plugin.video.reddit_viewer] 2.7.2

gedisony:
convert strings.xml to strings.po & cleanup unused strings
+misc updates


MartijnKaijser:
@gedisony could you please create a branch first and then commit your plugin?
Here's a little explanation
https://github.com/xbmc/repo-plugins/blob/master/CONTRIBUTING.md

PR #590, opened 2016-07-18
[plugin.video.reddit_viewer] 2.7.6

gedisony:


learningit:
This PR is a mess. You need to squash all the commits then do a compare and pull against the xbmc/plugins repo.

PR #591, opened 2016-07-19
[plugin.video.reddit_viewer] 2.7.6

gedisony:


learningit:
I'm pretty sure I requested that you change the strings.xml files to strings.po files in your next (this) update when I committed 2.7.5.

gedisony:
yes, no more strings.xml, just strings.po

learningit:
I didn't look at the default.py very carefully last PR, there's an awful lot of behaviour in there that would make this more of a script than a plugin - you actually have work arounds to accomplish some things that shouldn't be in a plugin.  Is there any particular reason that you made this a plugin?

gedisony:
Finnish and german files were in strings.xml. I removed them. 
This is my first addon, The original reddit addon for kodi have support for media from 4 or 5 sites. I decided to extend it. Do you want me to make this addon into script.something.something?

gedisony:
I can't get the plugin the work as a script. It is a plugin because the original was a plugin and I cannot find a workaround to make the plugin work as a script.

learningit:
I need to let someone who's more familiar with add-ons that combine/shift between a custom gui in a do.modal loop combined with plugin event management review this.  Some of the processing looks strange to me, but maybe it's ok.

PR #592, opened 2016-07-19
[plugin.video.espn_3] 2.9.14

awaters1:
Hello, we would like this new version of the ESPN 3 addon to be available in the official Kodi repository.  Thank you for the review.


plugin.video.espn_3/resources/lib/addon_util.py
```diff
@@ -0,0 +1,338 @@
...
+    video_streaminfo = dict()
+    liz.addStreamInfo('video', video_streaminfo)
+    ok = xbmcplugin.addDirectoryItem(handle=pluginhandle, url=u, listitem=liz)
+    return ok
+
+
+def addDir(name, url, iconimage="DefaultFolder.png", fanart=defaultfanart, infoLabels=None):
+    u = sys.argv[0] + '?' + urllib.urlencode(url)
+    xbmc.log(TAG + 'Made url to %s' % u, xbmc.LOGDEBUG)
+    liz = xbmcgui.ListItem(name, iconImage=iconimage, thumbnailImage=iconimage)
```
learningit: The inits for images are currently deprecated.  You should be using setArt() as you are below for fanart.

awaters1: Should be all set now, my eyes kept missing the deprecated message in the docs but I got it now.

PR #595, opened 2016-07-23
[plugin.video.manoto] 1.2.3

bbk79:


plugin.video.manoto/addon.xml
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8" standalone="yes"?>
 <addon id="plugin.video.manoto"
        name="Manoto TV"
-       version="1.2.2"
+       version="1.2.3"
        provider-name="babak">
   <requires>
     <import addon="xbmc.python" version="2.1.0"/>
```
learningit: Please change xbmc.python to 2.14.0 for gotham. Please see http://kodi.wiki/view/Addon.xml

plugin.video.manoto/default.py
```diff
@@ -34,29 +34,29 @@ def getStreamsFromPlayList(playlist):
         except urllib2.HTTPError, he:
         	print("HTTP error trying to open playlist")
```
learningit: Please remove the print statements from your modules. Please use xbmc.log or a support routine instead.

plugin.video.manoto/addon.xml
bbk79: Done!

plugin.video.manoto/default.py
bbk79: Removed

PR #598, opened 2016-07-27
[plugin.audio.cbcradio] 1.2.0

superphunthyme:
Kodi addon to listen to CBC radio stations online


learningit:
Can you replace the license.txt you have with GPL or similar?  Please see http://kodi.wiki/view/Add-on_rules
Your license.txt is a bit simplistic.

plugin.audio.cbcradio/addon.py
```diff
@@ -0,0 +1,121 @@
...
+    li = xbmcgui.ListItem('Radio 1 (' + my_r1_region + ')', iconImage='DefaultAudio.png')
+    li.setProperty('fanart_image',my_addon.getAddonInfo('fanart'))
+    xbmcplugin.addDirectoryItem(handle=addon_handle, url=url, listitem=li)
+
+    url = CBCJsonParser.parse_pls(CBCJsonParser.get_R2_streams(my_r2_region))
+    li = xbmcgui.ListItem('Radio 2 (' + my_r2_region + ')', iconImage='DefaultAudio.png')
+    li.setProperty('fanart_image',my_addon.getAddonInfo('fanart'))
+    xbmcplugin.addDirectoryItem(handle=addon_handle, url=url, listitem=li)
+
+    url = CBCJsonParser.parse_pls('http://playerservices.streamtheworld.com/pls/CBC_R3_WEB.pls')
```
learningit: This is sort of a stupid question, but are the streams from streamtheworld.com the original source for CBC radio? I can't find reference to them on the CBC websites (not sure I'm looking in the correct place though).

superphunthyme: They changed the website recently, so I don't know where they are now but those were the original links.
When I last tested them yesterday, they seemed to be working.
From [archive.org](https://web.archive.org/web/20140418014428/http://music.cbc.ca/#/streams.aspx)

superphunthyme:
For the license, I don't mind changing it to say the GPL.
For reference, the current one is the zlib license.

superphunthyme:
Other than that, anything to change? I'll add the GPLv3

learningit:
I believe either GPLv2 or GPLv3  is acceptable.

superphunthyme:
It looks like they've broken most of the streams with their redesign, except Radios 1 & 2.
It seems no longer possible to get the streams without a huge effort
I'll remove them but the addon is looking less and less useful.

PR #602, opened 2016-08-01
[plugin.video.euronews.com] 1.0.0

lambda81:
This is a new addon for euronews.
I decided to create a new one and not to update the old euronews addon for the following reasons:
1) The old one is broken for at least 1 year and it's not available on the official repo anymore.
2) The old one scraped the site, this addon is using the euronews api.


Tadly:
Looking good.
If there aren't any objection from @learningit or @phil65 I'd merge that...

PR #604, opened 2016-08-01
[plugin.video.mediathek] 0.7.2

raptor2101:
Changed:
- Kika is now "scrapped" by lxml.html
- ard interface changed addopted

Add:
- Italian translation


Tadly:
Your PR consists of two commits, one for `plugin.video.mediathek` (which this PR is actually about) and 
`plugin.video.gamestar` which is not supposed to be here.
Please get rid of the gamestar commit and do a separate PR for it.
Thanks

MartijnKaijser:
Nah this is fine like this. As long each is in it's own commit

Tadly:
@MartijnKaijser fair enough :)
/me will remember that

PR #600, opened 2016-07-31
Spurs TV 2.9.0

LS80:

PR #609, opened 2016-08-06
[plugin.video.spurs-tv] 2.10.0

LS80:
Addresses feedback from https://github.com/xbmc/repo-plugins/pull/600.


Tadly:
Looking good, thanks

LS80:
The version hasn't yet been updated in [addons.xml.gz](http://mirrors.kodi.tv/addons/krypton/addons.xml.gz). Is that supposed to have happened automatically after merge?

PR #617, opened 2016-08-16
[plugin.program.remote.control.browser] 1.0.4

chadparry:
This new add-on makes it easy to browse websites using a remote control instead of a mouse and keyboard. Here is the add-on description:

> The most powerful way to access content on Netflix and YouTube and Amazon Instant Video would be a web browser, if web browsers provided good native support for a 10-foot user interface. This add-on launches a browser and connects the arrow buttons on the remote control to the mouse pointer. This is the most user-friendly way to consume online content without needing a wireless keyboard.


learningit:
This really isn't a plugin, it should be a script. You start a bunch of background threads as well as a background httpserver and, it looks like, an LIRC daemon.  That is probably best done in a script.

There also seem to be a bunch of external dependencies which need to be installed for this to work correctly. I tried running and it just hung my system.

chadparry:
The way I understood plugins and scripts was that plugins can use scripting APIs and scripts can use plugin APIs, so there is not much difference. So is it just necessary for me to change the name and ID of this add-on from `plugin.program.remote.control.browser` to `script.remotecontrolbrowser`? I also have the extension point `xbmc.python.pluginsource` in my addon.xml file, and I think I need to leave that there, since I actually use the plugin GUI features.

> I tried running without them and it just hung my system.

None of the external dependencies should be able to hang your system. I'm concerned about that and I'd like to know more. Do you remember what you were doing when it hung? Did you have to restart Kodi, or reboot your system?

chadparry:
I did some more research and experiments on what could cause hanging. On a low-powered set-top box, I was able to reproduce a hang caused by Google Chrome. I worked around it by adding "--disable-gpu" to the command-line options. I'll incorporate that change into the next release.

I'm still curious what you would like changed so that this add-on can be accepted as a script.

phil65:
Is it possible to make the service part optional, or just start the service when someone opens the plugin part?
I think that´s what also irritates @learningit 	, a user doesnt really expect to install some always-running service when installing a program.

chadparry:
I agree with that sentiment. The HTTP server is already off by default. The user has to go to the add-on settings to enable it if they want it to run. The browser and the other background threads only run while the plugin is being used. Once the user is done browsing websites, all those threads are killed. There is virtually no overhead while the user is doing anything else in Kodi. The only thread that sticks around is listening on `xbmc.Monitor.waitForAbort` and `xbmc.Monitor.onSettingsChanged`, so it knows when the add-on config needs to be updated.

phil65:
Ok. If that part is optional and opt-in and you make use of VFS browsing then I think it´s fine to classify this as a program plugin (it will appear in the same section as scripts anyways, "programs".)

learningit:
Actually what I was thinking is that when run as a plugin the threads can be running in the background while other things are running and have the potential of becoming zombies which present some issues when exiting Kodi.  For example you just need to hit the back key when running as a plugin to throw it into the background and navigate elsewhere while the plugin remains running.

I've found that I needed to run as a modal script that cleans everything up when you exit the do.modal loop.  Maybe I'm wrong and things have changed in newer versions of Kodi.

plugin.program.remote.control.browser/addon.xml
```diff
@@ -0,0 +1,32 @@
...
+        <disclaimer lang="pt_BR">A experiência será degradada, a menos que essas dependências externas estejam instaladas: “psutil”, “pyalsaaudio”, “pylirc2” e “Pillow”. Outra ferramenta útil é “unclutter”, que oculta automaticamente o ponteiro do mouse. (Em um sistema baseado em Debian, execute “sudo apt-get install python-psutil python-alsaaudio python-pylirc python-pil unclutter”). Finalmente, um tema com um ponteiro do mouse grande irá melhorar a visibilidade do ponteiro, (por exemplo, https://www.gnome-look.org/p/999574/).</disclaimer>
+        <platform>linux osx</platform>
+        <language>en en_US de_DE pt_BR</language>
```
phil65: I think this one is only needed in case the plugin content is localized and ,because of that, not really useful to english speakers for example. So it doesnt really make sense to use it in this case.
Is that correct, @learningit ?

phil65:
@learningit 	Hmm could you elaborate? Do you have any more concrete examples? I dont really see any problems, but perhaps i am missing something.

chadparry:
@learningit, I had tested this add-on by switching from the browser back to Kodi and then quitting Kodi. The thing that I found worked the best was to have a thread [listening to `xbmc.Monitor.waitForAbort`](https://github.com/xbmc/repo-plugins/pull/617/files#diff-9d636221cd8f9439c20761d1c1d30cf9R394) as I mentioned before, and to kill the browser and all other threads when Kodi exits. It presents a clean UX with no zombie processes. If I didn't do that, then the running plugin would have prevented Kodi from exiting completely. I can also experiment with a modal window, but I think the current implementation does a pretty good job of providing the behavior that you are asking for.

plugin.program.remote.control.browser/addon.xml
chadparry: Do you mean the `language` element? All the plugin content is actually localized to four languages (if you count American English and British English as separate).

phil65: This is not about the languages the plugin labels are translated to, but about the language of the video content.
I think we only use it for non-english content. @MartijnKaijser ?

chadparry:
Ping. This is ready to be accepted, right?

learningit:
You need to squash the commits to a single commit and name it properly. Please see: https://github.com/ginatrapani/todo.txt-android/wiki/Squash-All-Commits-Related-to-a-Single-Issue-into-a-Single-Commit

chadparry:
Ping! Is there anyone who can approve this pull request? Thx.

phil65:
As already mentioned, please leave the language element in addon.xml empty.

plugin.program.remote.control.browser/addon.xml
enen92: I think that's correct. At least when I first submitted the RTP Play addon I also made the same mistake (english and portuguese as language options due to the addon available translations). I was advised by Martijn (IIRC) to change it to Portuguese only since it was only about the video content.

phil65: Thx for confirmation. @chadparry 	, could you make this change? After that this is good to go  I think.

plugin.program.remote.control.browser/addon.xml
```diff
@@ -0,0 +1,31 @@
...
+        <description lang="pt_BR">A forma mais poderosa de acessar o conteúdo no Netflix e YouTube e Amazon Instant Video seria um navegador web, se navegadores web fornecessem um bom suporte nativo para a interface de usuário de 10-foot. Este add-on inicia um navegador e conecta os botões de navigação do controle remoto para o ponteiro do mouse. Esta é a interface mais amigável para consumir conteúdo online sem a necessidade de um teclado sem fio.</description>
+        <disclaimer lang="en">The experience will be degraded unless these external dependencies are installed: “psutil,” “pyalsaaudio,” “pylirc2,” and “Pillow.” Another helpful utility is “unclutter,” which automatically hides the mouse pointer. (On a Debian-based system, run “sudo apt-get install python-psutil python-alsaaudio python-pylirc python-pil unclutter”). Finally, a theme with a large mouse pointer will improve pointer visibility, (e.g., https://www.gnome-look.org/p/999574/).</disclaimer>
```
phil65: same here.

plugin.program.remote.control.browser/addon.xml
```diff
@@ -0,0 +1,31 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<addon id="plugin.program.remote.control.browser" name="Remote Control Browser" version="1.0.2" provider-name="Chad Parry">
+    <requires>
+        <import addon="xbmc.python" version="2.19.0"/>
```
phil65: ..and change this to 2.24.0

plugin.program.remote.control.browser/import.py
```diff
@@ -0,0 +1,2 @@
+import PIL.Image
```
phil65: if you want to use PIL then you need to add it to your addon.xml dependencies.

plugin.program.remote.control.browser/plugin.py
```diff
@@ -0,0 +1,990 @@
...
+                    proc.wait()
+                    proc = None
+                    xbmc.log('Waited for the browser to quit', xbmc.LOGDEBUG)
+                finally:
+                    terminator.cancel()
+                    terminator.join()
+            finally:
+                if proc is not None:
+                    # As a last resort, forcibly kill the browser.
+                    xbmc.log('Forcefully killing the browser', xbmc.LOGINFO)
```
phil65: please always use LOGDEBUG.

plugin.program.remote.control.browser/plugin.py
```diff
@@ -0,0 +1,990 @@
...
+        isAborting.set()
+        if proc.poll() is None:
+            proc.kill()
+        proc.wait()
+        if activator is not None:
+            activator.join()
+
+
+@contextlib.contextmanager
+def lockPidfile(browserLockPath, pid):
```
phil65: hmm not the biggest fan of this.... what do you need it for exactly?

plugin.program.remote.control.browser/resources/language/English/strings.po
```diff
@@ -0,0 +1,132 @@
+# XBMC Media Center language file
```
phil65: would be cool if you could use the new folder naming scheme for language folders (in this case resource.language.en_gb )

plugin.program.remote.control.browser/service.py
```diff
@@ -0,0 +1,432 @@
...
+    def __init__(self, addon):
+        super(LinkcastMonitor, self).__init__()
+        self.addon = addon
+
+    def onSettingsChanged(self):
+        self.addon.reloadLinkcastServer()
+
+
+class LinkcastServer(BaseHTTPServer.HTTPServer):
+
```
phil65: A small explanation please why you need a HTTPServer?

plugin.program.remote.control.browser/service.py
```diff
@@ -0,0 +1,432 @@
...
+        return BOOL_ENCODING[bool(val)]
+
+    def unmarshalBool(self, val):
+        STRING_ENCODING = {'false': False, 'true': True}
+        unmarshalled = STRING_ENCODING.get(val)
+        if unmarshalled is None:
+            raise ValueError('Invalid Boolean: ' + str(val))
+        return unmarshalled
+
+    def isPillowInstalled(self):
```
phil65: pillow is available on all platforms and can get imported via addon.xml.

phil65:
Sorry for the late review and thanks for your patience. I think my review is finished now. :)

learningit:
Ok. So when the above changes requested by phil65 are commited, I'll merge. Thanls

plugin.program.remote.control.browser/plugin.py
chadparry: Isn't LOGINFO used for extreme conditions that the user would want to be aware of?

plugin.program.remote.control.browser/plugin.py
chadparry: The user can trigger a browser to start either though the Kodi GUI or through the remote bookmarklet API. But with multiple start mechanisms, they need to make sure they don't clobber each other. If the plugin is already running, then other threads that try to start the browser will just fail. The pidfile is the locking mechanism.

plugin.program.remote.control.browser/resources/language/English/strings.po
chadparry: Would the new folders be backwards compatible with Jarvis?

plugin.program.remote.control.browser/service.py
chadparry: Excellent! I wish I had known that.

chadparry:
I'll prepare another minor version release with the changes.

plugin.program.remote.control.browser/resources/language/English/strings.po
phil65: yes.

plugin.program.remote.control.browser/plugin.py
phil65: Hmm I might be wrong, not 100% sure. @learningit do we have a policy here?

phil65:
Cool, thx. Perhaps wait for outstanding comments before updating the PR again.

plugin.program.remote.control.browser/plugin.py
phil65: @learningit, fine for us?

learningit: I'm not a fan of this either. Perhaps a window property?

plugin.program.remote.control.browser/plugin.py
phil65: ok. this is fine then.

phil65:
I think you can prepare the update now. All reviewed.

phil65:
bump.

chadparry:
I've incorporated all the suggestions from this review. I was already in the process of making some other enhancements, and I'll have to test those a little longer before my next version release will be ready.

phil65:
ok, cool. we will wait then.

chadparry:
I've committed all the changes that were requested. Let me know if you need anything else. Thanks!

chadparry:
Ping… This pull request is all ready to be merged. Thanks!

chadparry:
Ping… Who is able to merge this pull request?

Tadly:
First of, I'm sorry this PR is tacking its sweet time but this is quite a special case.

I would have an alternative for the settings.xml generation (so you don't have to do it at all, unless there is more to it than the "dynamic" visibility of settings)
fixing that issue **but** the whole pid-tracking / process killing (no matter how good you implemented it) is something we (or I at least) can't really accept.

Not in the official repository at least.

I'm really sorry, especially for all the time you've spend trying to get it in the repository and I hope you understand the reasoning for this decision.

If you're interested in how you could change the settings generation thing, lemme know and I'll gladly throw an example in the comments (still a not fully developed idea but should work.. I think)

chadparry:
@Tadly, of course I would prefer to get this in the official repository. If you want the settings done differently, then please let me know what you would like done. I just want to generate defaults and warnings that are appropriate for the platform. And if you don't like process management, then please let me know why. Of course this functionality could be approximated without killing the browser process, but the quality would probably go down. I'm happy to discuss it.

Tadly:
To display the warnings you could do something like this...
```
<?xml version="1.0" encoding="UTF-8"?>
<settings>
    <!-- You set these to either true or false from your service.py -->
    <setting id="has_psutil"    type="bool" visible="false" default="false" />
    <setting id="has_alsaaudio" type="bool" visible="false" default="false" />
    <setting id="has_pylirc"    type="bool" visible="false" default="false" />

    <!-- These warnings are only visible if their counterparts are false -->
    <setting label="Warning: psutils not found"   type="lsep" visible="eq(-3,false)" />
    <setting label="Warning: alsaaudio not found" type="lsep" visible="eq(-3,false)" />
    <setting label="Warning: pylirc not found"    type="lsep" visible="eq(-3,false)" />
</settings>
```
On to those "Browser defaults". Don't use the settings "default" feature and instead set the value to whatever you generated **if** there isn't anything set yet.

After that, I think, there's no real need to generate a whole _settings.xml_ from within your source (unless I missed something).

> Of course this functionality could be approximated without killing the browser process, but the quality would probably go down.
We don't wanna loose quality. :)
I raised some questions internally and will come back to you after I got em answered.

chadparry:
@Tadly, that's a great suggestion to change the settings values but keep the settings schema intact. I've published "release-1.0.5" of my addon with that change included. It worked beautifully. Should I create another pull request for it?

As for the process management, I realize that it is unusual for a Kodi add-on. This add-on has the unique requirement that it will start and stop a real browser, such as Chrome. The strange part really is only that an external browser process is started, and once that is accepted as a requirement then the process management requirement follows naturally. This is hands-down the best way to browse Netflix and Amazon Instant Video on a 10-foot display. The PID tracking is useful for the following scenarios. Many of them could be removed, but the current behavior is quite a good user experience.

1. When the user hits the `BACK` button on the remote control, that triggers the browser to exit. This is done by sending it a `SIGTERM`, which is a supported way to exit cleanly. (As a fail-safe, the browser process's children get signaled too, because someone might specify a custom shell script that starts the browser, and both should be killed). (For additional robustness, the browser is sent a `SIGKILL` if it doesn't exit within a reasonable timeout).
1. Similarly, if the user tries to exit from Kodi while the add-on is running, then the add-on detects that and requests the browser to exit also.
1. When the user exits from the browser manually, such as with `ALT-F4`, the add-on detects that and shuts itself down cleanly, handing control back to Kodi.
1. The browser process is raised to the foreground once it starts up. That is a trick that was also employed by the `plugin.program.chrome.launcher` add-on, and it seemed to be helpful on some platforms.
1. The PID is used to create a pidfile within the add-on, which prevents multiple browsers from being started at the same time.

Please let me know if you have any questions or suggestions.

Tadly:
Thanks for taking the time to explain all this is more detail.

I'd actually like to merge the add-on after all (the new PR with reworked settings that is) *unless* the others are against it.

@phil65 @learningit  what do you say?

Tadly:
@chadparry eyup. Discussed it internally again.
You can create a new PR (with the settings re-work) and I'll merge that one

chadparry:
@Tadly, thanks! I've submitted https://github.com/xbmc/repo-plugins/pull/775.

PR #624, opened 2016-08-18
[plugin.video.reddit_viewer] 2.7.9

gedisony:
Removed custom guis and workarounds.


learningit:
In default.py where you instantiate a list item, the initializers for iconImage and thumbnailImage are deprecated, i.e.
liz = xbmcgui.ListItem(name, iconImage="DefaultFolder.png", thumbnailImage=iconimage)

Please use ListItem.setArt() instead.  Also please set 'mediatype' to the correct value for the ListItem.

learningit:
Because you have a pretty large default.py, you might want to consider a stub default.py in the root addon directory that instantiates a class from a .py file in resources/lib or does an include, that way only the stub is compiled on load, the the .py in resources/lib will only be compiled when it changes, the add-on should load much faster.  Just a suggestion.

PR #627, opened 2016-08-19
[plugin.program.flypdx] 0.0.5

troypdx:
This plugin displays flight arrival/departure information for the Portland International Airport (PDX).

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,95 @@
+### Introduction
+
+**Kodi** uses Github for development only, i.e. for *pull requests* and the review of such code.  
+**Do not open** an issue on Github for your questions or bug reports.  
+**Do not comment** on a *pull request* unless you are involved in the testing of such or have something meaningful to contribute.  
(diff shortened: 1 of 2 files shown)


Tadly:
Hi,
There're a few things wrong with your PR.
 1. Your PR should consist of only **one** commit which includes all changes (or in your case, as it is a new submission, the whole source). Look at other, open/closed PRs to get an idea.
 2. your addon.xml uses xbmc.python version 2.1.0 which is [deprecated](http://kodi.wiki/view/addon.xml#Dependency_versions). As you submitted to the Jarvis repo, you might as well set it to 2.24.0
 3. Looking at the repository on your profile, there's a `requests` folder (which I assume is supposed to be a git submodule). That can and should be removed as you specified requests as a dependency in your addon.xml.  
Talking about dependencies, you don't use any of those so you should remove them all together from your addon.xml.

Once you got through all that, I'll have a proper look at the source.
Thanks for your submission though :)

phil65:
Is a plugin for a single airport worth it?

PR #631, opened 2016-08-23
plugin.audio.calmradio - v0.2.4

nire0510:
verify flag added to requests.get method


Tadly:
Doesn't look like you even use your change anywhere.
I'll merge it for now, but make sure to only PR changes that are necessary (fixes, new features etc.).
Thanks ;)

nire0510:
thanks @Tadly, but I do use this change - need to ignore SSL certificate verification which is set to True by default.

Tadly:
Whops, completely ignored the fact that you set a default value for that argument.
Totally my bad, I apologize :)

PR #632, opened 2016-08-24
[plugin.video.tyt] 2.0.0

lottaboost:
Updated plugin.video.tyt to version 2.0.0, major update.  Allows non-members and members to access all free content.


lottaboost:
I know I'm supposed to squash my commits into one, but it won't allow me.  When I do a "git status" I get this - 
[status.txt](https://github.com/xbmc/repo-plugins/files/434636/status.txt)
then if I do a "git diff" i get this
[diff.txt](https://github.com/xbmc/repo-plugins/files/434642/diff.txt)
The reason for all the diffs in each file are "The file will have its original line endings in your working directory.
warning: CRLF will be replaced by LF in plugin......"
This error can be removed normally by editing the .gitattributes and commenting out the text=auto.  I think the reason for this error is the writers of the plugins is that the actual kodi repo has files that are using CRLF's instead of LF, maybe the people that wrote those plugins were written on a windows system?  http://stackoverflow.com/questions/1967370/git-replacing-lf-with-crlf
I don't believe the error is from something on my end, as I run linux.

lottaboost:
Sorry.  I thought that once a plugin was merged, you could  update whenever.  This is going to be the last update for a while, as I have all the content accessible that is relevent to the plugin.

PR #641, opened 2016-08-30
[plugin.video.tyt] 2.0.0

lottaboost:
Made plugin for non-members also.  Added tons of content, including live stream and youtube channels.


plugin.video.tyt/resources/language/English/strings.po
```diff
@@ -1,15 +1,22 @@
 msgid ""
 msgstr ""
 
-msgctxt "#3000"
+msgctxt "#30000"
```
phil65: please change to id range 32000-33000

phil65:
Would be nice if you could change the string ids.
Apart from that it looks fine to me.

plugin.video.tyt/resources/language/English/strings.po
ronie: i don't claim to understand the difference (and it would appreciated if someone could enlighten me), but plugins should use the 30000-31000 range :
https://github.com/xbmc/xbmc/blob/master/addons/resource.language.en_gb/resources/strings.po#L15292

lottaboost:
I used the string range that was specified in http://kodi.wiki/view/Add-on_structure.  According to the wiki the range that you want me to use 32000-33000 are reserved for scripts.  The range I used, according to the wiki, are for "plugins and plugin settings", and are what I've seen used for all the video addons that have been used on the Kodi repo.  Also, 33000-33999 are reserved for "common strings used in addons".  All the video a

phil65:
oh, you might be right, I thaught any add-on should use 32000-33000.
Using 30000-31000 is fine for plugins, @ronie?

tamland:
The range thing is just legacy. Except for skins they have no meaning. You can use whatever numbers you want..

phil65:
cool, thx for clarification, @tamland

lottaboost:
Thanks for the clarification.  My previous merge, I used 3000 range (Forgot a zero), that's why I changed it.

lottaboost:
Ok, I'll leave it as it is.

PR #652, opened 2016-09-04
[plugin.video.funimationnow] 0.0.3

Protocol-X:


learningit:
Please squash the commits and rename the resulting commit in the same format as the PR., i.e. plugin.video.blah x.y.z

Code looks good, you may want to consider using xbmc.waitForAbort() rather than the sleep calls in player.py, Also a bit confusing that a couple calls use control.sleep and others use xbmc.sleep - looks like they're the same.

Protocol-X:
Thanks I will take a look.  I did run the commands to squash but it still
showed 2 commits.  I will try again and then get back to you.

learningit:
Just noticed there is an existing funimation add-on in the kodi.org repo. Are you the original author there as well? It's fairly new, though marked broken, so waht to make sure that the original author is ok woth this. At any rate, you need to bump the version beyond 1.2.1

learningit:
You need to do a force push to the branch to update the squash. Please see: https://github.com/ginatrapani/todo.txt-android/wiki/Squash-All-Commits-Related-to-a-Single-Issue-into-a-Single-Commit

Protocol-X:
I have already discussed this with another member.  Funimation and
funimationnow are different.  The names look simular so it can be
overlooked easily

learningit:
Ok, see it now. I was confused because the icon just says "Funimation" without the "now".

Protocol-X:
This is the method I used bit it didn't work.  I am at the office now so I
have no access to try again. I will update it later tonight when I get home
and let you know.  Appreciate the feedback

Protocol-X:
I have ran through this 2 more times and even know there are no errors it always shows 2 commits

Protocol-X:
Please see new request https://github.com/xbmc/repo-plugins/pull/665

PR #653, opened 2016-09-04
[plugin.video.frikanalen] v0.0.1

scanf:


phil65:
imo this does not qualify as a plugin.... it´s just one single entry.

scanf:
Sorry, I was not aware of a minimum numbers of entries to qualify as a plugin.

Our motivation to get our changes merged was to ease setup for people who want
to watch Frikanalen. Making it a Kodi plugin, makes it easily available IMO.

Not sure if you have run the plugin, but the entry let's you stream the channel. Is there
any change we can do to qualify as a plugin? Or should I close this?

Thanks.

tamland:
Sorry @phil65, I'll have to overrule you on this one;) In this case it should be fine. It would be *nice* to have recordings here as well, but afaik we don't have any requirements/rules for it.

A few things that must be fixed though:
-description says "NRK". surely a copy-paste error (?)
-script.module.requests and script.module.routing are listed as dependencies but not used
-please follow the icon guidelines here: http://kodi.wiki/view/Add-on_structure#icon.png

phil65:
sure sure. :) since the VFS functionality is not really used here i was just wondering if this wouldnt be better suited as a script.

scanf:
The issues should now be addressed. Specifically the description is now more
consistent and the unused modules have been removed.

The only extra change that was not mentioned was the removal of the svg file.
Hope that does not affect the review.

Thanks.

scanf:
Made the list item playable and the mediatype is now set. I am doing this for the first time, so I might have made mistakes.

plugin.video.frikanalen/addon.xml
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<addon id="plugin.video.frikanalen"
+       name="Frikanalen Nett-TV"
+       version="0.0.1"
+       provider-name="frikanalen">
+  <requires>
+    <import addon="xbmc.python" version="2.1.0"/>
```
phil65: this should be 2.24.0

plugin.video.frikanalen/default.py
```diff
@@ -0,0 +1,35 @@
...
+
+streamurl = 'http://video.nuug.no/frikanalen.webm'
+
+import sys
+import xbmcgui
+import xbmcplugin
+
+addon_handle = int(sys.argv[1])
+
+xbmcplugin.setContent(addon_handle, 'movies')
```
phil65: please set content to "videos" or "files"

scanf:
Changed version from 2.1.0 to 2.24.0 and content is now set to videos.

Thanks for the patience.

PR #661, opened 2016-09-11
[plugin.video.vice.com] 1.0.0

lambda81:
This is an alternative addon for vice.
It gets videos from vice directly, not from the vice youtube channel.
It also supports the selection of vice videos based on country.
It also has subtitle support (based on selected country).


learningit:
We only allow one add-on per source in the repo. There is already an add-on for vice.com in the repo and it's relatively up to date.  You may want to contact the author in the repo to collaborate on a joint update.

lambda81:
technically it's not the same source

this addon is for
http://www.vice.com/videos

the other addon is for
https://www.youtube.com/user/vice

thank you for your quick reply and feedback.

phil65:
is it just technically different or are the videos also different?

lambda81:
the videos are different, vice only uploads selected videos on youtube.
this addon includes videos on different languages too.

available countries are Australia, Austria, Belgium, Brazil, Canada, Czech Republic, Colombia, Denmark, France, Germany, Greece, Italy, Mexico, Netherlands, Poland, Portugal, Romania, Russia, Serbia, Spain, Sweden, United Kingdom and United States

some examples
http://www.vice.com/fr/videos
http://www.vice.com/gr/videos

there are only english videos on youtube.

lambda81:
this is more close to the plex vice addon
https://github.com/plexinc-plugins/Vice.bundle

DamonToumbourou:
I confirm that I accept changes.

PR #662, opened 2016-09-14
[plugin.video.vice] 1.1.0

lambda81:
This is a collaboration on a joint update of vice addon.

https://github.com/xbmc/repo-plugins/pull/661#issuecomment-246178916

Damon will confirm it.


learningit:
Just a tip: Looking through vice.py, it's not great practice to wrap things in try/except, especially when it's nested. You probably could have avoided a lot of the try/except by using a = x.get(n, default) rather than try:  a=x[n], particularly visible in the tvtitle and duration routines.  I started the same as you with lots of try/excepts and @martijn and @phil65 pointed out these issues and the better way with a .get(), I've never gone back.

lambda81:
I understand what you are saying, I will remove them on future updates

PR #681, opened 2016-10-01
[plugin.program.video.node.editor] 1.0.2 (Mark broken)

Ignoble61:
Based on the Library Node Editor replacing this script from Isengard onwards, this marks the script broken so that it is no longer available to install in favour of the maintained script.

Ping @ronie to confirm that the Library Node Editor was introduced with Isengard.


ronie:
this should be good to go

PR #726, opened 2016-10-26
[plugin.video.snnow] 1.0.7

micahg:
Attempts reauthentication when stream retrieval fails and fixes sportsnet authentication.


micahg:
TYVM.

PR #737, opened 2016-11-04
[plugin.video.fsgo] 1.1.3

emilsvennesson:


plugin.video.fsgo/default.py
```diff
@@ -50,7 +52,7 @@ def addon_log(string):
         xbmc.log(msg=msg, level=xbmc.LOGDEBUG)
```
phil65: xbmc.LOGDEBUG should already make sure that log info only gets printed for debug mode. Why do you check for debug mode  via JSON?

emilsvennesson: @phil65 fsgo.py is Kodi independent and requires the 'debug' parameter so I can't really rely on xbmc.LOGDEBUG. I could lie and send True, but then the debug code (which can be a bit heavy in some cases) would consistently even if it's not getting printed in the log file. 

This feels like the best solution IMO.

phil65: ok, thx for explanation. Forgot that the module was kodi-independent.

emilsvennesson: @phil65 Thanks for the review. :) This is probably the wrong place for feature requests, but it would be nice if you could access this setting using xbmc.getCondVisibility.

PR #744, opened 2016-11-06
[plugin.audio.pureradio] 1.0.0 

bdommel:
### Description
This plugin enables the playback of the PureRadio audio stream, which can be set in different qualities using different bitrates. Also a selection of podcast, hosted on Spreaker, are available for instant playback.
### Checklist:
- [X ] My code follows the [add-on rules](http://kodi.wiki/view/Add-on_rules) and [piracy stance](http://kodi.wiki/view/Official:Forum_rules#Piracy_Policy) of this project. 
- [ X] I have read the [CONTRIBUTING](https://github.com/xbmc/repo-plugins/blob/master/CONTRIBUTING.md) document
- [X ] Each add-on submission should be a single commit with using the following style: [plugin.video.foo] v1.0.0


Razzeee:
The first commit tries to add the folder ` plugin.audio.pureradio.git` I think that's not what you want to do is it?

bdommel:
I thought I had that removed and cleaned it up before commiting. It should only contain the plugin folder.

bdommel:
I removed the folder with the .git extention. Sorry for the inconvience.

Razzeee:
Are you able to squash these three commits into one?

bdommel:
I squashed the commits.

bdommel:
Perhaps it is better to close this commit and do a single commit with all changes.

PR #757, opened 2016-11-11
[plugin.video.mediathek] 0.7.5 (depcrate pre krypton version)

raptor2101:
### Description
ZDF Completly changes its WebInterface so the plugin is rewriten to adopt that (now using JSON API, hopefully more stable)
ZDF enforces TLSv1 with SNI enabled. Support for this is added in Pyhton 2.7.9. (See here https://github.com/raptor2101/Mediathek/issues/85). Therefore i mark it as broken. 
### Checklist:

### DIFF
--- a/plugin.video.mediathek/addon.xml
+++ b/plugin.video.mediathek/addon.xml
@@ -36,6 +36,6 @@ Al momento sono supportate le seguenti emittenti:
     <website>https://github.com/raptor2101/Mediathek</website>
     <source>https://github.com/raptor2101/Mediathek</source>
     <email>raptor2101@gmx.de</email>
-    <broken>deprecated</broken>
+    <broken>Because of rollout of TLSv1 with SNI enabled, Python 2.7.9 (and later) is needed. Please refere to the changelog for more information.</broken>
   </extension>
 </addon>
--- a/plugin.video.mediathek/addon.xml
+++ b/plugin.video.mediathek/addon.xml
@@ -2,7 +2,7 @@
 <addon
   id="plugin.video.mediathek"
   name="Mediathek"
-  version="0.7.4"
+  version="0.7.5"
   provider-name="Raptor 2101">
   <requires>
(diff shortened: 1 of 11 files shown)


Razzeee:
You can basically only update `addons.xml` and the changelog. Drop every other change, as marking as broken will just lead to those users deinstalling the plugin.

raptor2101:
Is this a won't be pulled this way? My intention was, to supply the last functional version ...

phil65:
not sure how we handle deprecation... @MartijnKaijser ?

Rondom:
The plugin works perfectly fine on older versions if you have a newer Python (e.g. on Linux). Even with an old Python, it will still work fine for the other TV stations supported.

Would a check in the relevant code that emits a warning under older Python-versions be an alternative?

raptor2101:
@Rondom: this is not realy an option. TLS with active SNI gets deployed more frequent. (Currently ARTE are the next one)

I have no setup (most error reportings comes from open elec users) to distinguish between errors caused by broken TLS implementation and real "Parsing errors" (eg my pluging engaged a wrong url)

@phil65: thanks for your attention :+1:

MartijnKaijser:
@raptor2101 you might want to make the broken message a bit more descriptive like "Upgrade to new Kodi v17 to keep using this plugin".

Problem now is that the version in Krypton repo has same addon version and that won't work as it needs to e higher to override this deprecated version for v17 users. So bump that one first and then we can merge this

raptor2101:
@MartijnKaijser: Version bumb done - How to present a "broken message". Simply write some string in the broken tag?

MartijnKaijser:
@raptor2101 yep. Whatever is between the tags should be shown to the user if he looks at the addon in repo manager. Initially he will just get a "addon has been marked broken" message.

raptor2101:
@MartijnKaijser Message is now more "helpful" (hopefully) :)

PR #771, opened 2016-11-21
[plugin.video.youtube] 5.3.5

jdf76:
Updating youtube plugin from 5.2.1 to 5.3.4


plugin.video.youtube/resources/lib/youtube/helper/subtitles.py
```diff
@@ -0,0 +1,182 @@
...
+
+    def subtitle_url(self, sub_language):
+        return self.SUBTITLE_URL % (self.video_id, sub_language)
+
+    def _languages(self):
+        return self.context.get_settings().subtitle_languages()
+
+    def _write_file(self, _file, contents):
+        self.context.log_debug('Writing subtitle file: %s' % _file)
+        f = xbmcvfs.File(_file, 'w')
```
phil65: please use "with" statement for opening files.

plugin.video.youtube/resources/lib/youtube/helper/yt_specials.py
```diff
@@ -51,17 +56,22 @@ def _process_browse_channels(provider, context, re_match):
...
+        if not v3.handle_error(provider, context, json_data):
+            return False
         result.extend(v3.response_to_items(provider, context, json_data))
         pass
 
     return result
 
 
 def _process_disliked_videos(provider, context, re_match):
+    provider.set_content_type(context, kodion.constants.content_type.EPISODES)
```
anxdpanic: This would change the whole view system?

plugin.video.youtube/resources/lib/youtube/helper/subtitles.py
anxdpanic: with is supported here? I went by docs for xbmcvfs will correct when I get a chance.

plugin.video.youtube/resources/lib/youtube/helper/yt_specials.py
phil65: Which view system?
It´s easy: Youtube videos are not episodes, they are videos.

plugin.video.youtube/resources/lib/youtube/helper/subtitles.py
phil65: oh, sorry, i think youre correct. sorry for the noise. :) some try-except error handling would still be nice though when dealing with files.

plugin.video.youtube/resources/lib/kodion/impl/xbmc/xbmc_context.py
```diff
@@ -229,4 +230,36 @@ def sleep(self, milli_seconds):
...
+        try:
+            return response['result']['addon']['enabled'] == 'true'
+        except KeyError:
+            message = response['error']['message']
+            code = response['error']['code']
+            error = 'Requested |%s| received error |%s| and code: |%s|' % (rpc_request, message, code)
+            xbmc.log(error, xbmc.LOGDEBUG)
+            return False
+
+    def set_addon_enabled(self, addon_id, enabled=True):
```
phil65: that would be nice, thx.

plugin.video.youtube/resources/lib/youtube/helper/yt_specials.py
anxdpanic: The override view that is configured during the setup wizard if a supported skin is installed or in settings. It's been there a long time. As I understood it, different views are available for different content types, episodes fits as far as views / youtube information per video for the most part which is why I'm guessing this is what was chosen. This change would cancel out that option for all these users that are used to it. To keep the option with this change would require gathering of view id per skin for video content type which may not have the view users are now accustomed to. If you need it changed, I don't understand but I will change it.

anxdpanic: Based on your request, and reading http://i.imgur.com/TeXMOMx.png from (https://kodi.tv/kodi-v17-krypton-default-skin-next-gen/) I have removed the override views and set them all to videos. It is limited in view selection especially on < Krypton so I expect this will not be something users are going to like.

plugin.video.youtube/resources/lib/kodion/impl/xbmc/xbmc_context.py
anxdpanic: This as well as try/except for xbmcvfs file write has been addressed.

plugin.video.youtube/resources/lib/youtube/helper/yt_specials.py
phil65: > episodes fits as far as views / youtube information per video for the most part which is why I'm guessing this is what was chosen.

Plug-in authors shouldnt workaround the "issue" that some skins do not pay attention that much to the more seldom used content types. The whole point of our plug-in system is that the plug-in author should not need to know about any skin´s inner workings. These workarounds just hide the problems we need to solve in a proper way.
By using "episodes" as content type, several skins will display factually wrong information at some places. (Example: Lot of skins show "TV shows" in breadcrumbs when content is set to "episodes".)

phil65:
Thx for the adjustments.
If some people are complaining about the viewtype changes, please point them to me. ;)

jdf76:
Phil,

Quick question. I noticed that when I use the new version I’m limited to 2 views on the list. Shift and list. When I had it before it had more options. I”m assuming it’s due to it being labeled as episode before, so my question is, do you mind if we release an alternative unsupported version that puts back the old options to make it more like the older version? It won’t be listed in a repo, just a link on the forum so people to manually download. 

Do you guys have any issues with this?

Let me know. 
Thanks
Jeff

phil65:
Skins need to allow Container.Content(videos) for viewtpes that are appropriate for this kind of media.
I hope that I will find some time to write some explanation text in dev section which container.content types should be used by plug-in authors and which content types should be dealt with by skin authors.

plugin.video.youtube/resources/lib/kodion/impl/xbmc/xbmc_context.py
MartijnKaijser: only inputstream.adaptive (included since v17 beta6) will remain as inputstream.mpd will vanish before release.

anxdpanic: 👍  the change to inputstream.adaptive has already been forced pushed

anxdpanic:
not sure what happened, the current helix branch code and resulting zip do not match this PR

PR #779, opened 2016-12-03
[plugin.video.youtube] 5.3.6

anxdpanic:
resolving #771 PR <-> resulting commit mismatch causing broken add-on.


MartijnKaijser:
Btw most videos are set as episodes while they are not. As such it will add tagging that should not be there. 
Maybe @phil65 can give some hints

anxdpanic:
the correction for mediatype has been commited https://github.com/Kolifanes/plugin.video.youtube/commit/78cf3c25de2b535d15be19c35b83ff06869a7b4b
got focused on correcting content type, override views and dash that it was overlooked. 

Anything >720p (non-live) is either separate audio/video streams or dash. For setting 4K I believe you need a 4K resolution, and in inputstream.adaptive leave quality to auto and min bandwidth to a number higher than highest required bandwidth of a 4K stream. Then it's more a forced 'highest quality'.

MartijnKaijser:
I have everything 4k and inputstream enabled ;) 
I meant in YouTube settings you have the resolution setting which now is maximum 1080p/720p  and other lower options. Now I expect it to max it out to 1080p even though there could be higher streams.

anxdpanic:
I assumed you did :) Caused the question confuse me a bit hah. 

1080 was the highest that I recall ever seeing myself. The 1080 label was changed so it doesn't set an expectation of 1080 content from videos only live content (still functions the same). I haven't seen any examples of >720p video content, which has been the reason for all the 'when is dash coming' and 'is there 1080 yet' questions. 1080 would have worked this whole time had it been available, but >720 is only available as separate audio/video streams or dash that I've seen.

I'd expect to see less available this way in the future not more, but that is not based on any information from Google/YouTube either.

As far as adding a 4K enum, it _could_ be done but to me seems more like setting an expectation that it will be available in a case where it's slim we'll ever even see 1080.

MartijnKaijser:
Well from testing and setting inputstream to manually select I get 4k video. I'll see how this will work out and if inputstream needs to be adjusted

PR #803, opened 2016-12-17
[plugin.video.rtlxl] 1.1.5

Opvolger:
[plugin.video.rtlxl] 1.1.5


Opvolger:
Sorry, i didn't see my mistake.
I have now already two emails from people how have this problem.

This should fix it.

PR #931, opened 2017-01-29
[plugin.video.snnow] 1.17.11

micahg:
### Description
HLS is broken for authenticated Neulion streams on Krypton -- I think it has to do with FFMPEG loading a key on each stream while it probes... I haven't had a chance to look into a proper fix in Kodi just yet (not sure I'll have time) but to minimise complaints I'm working around it by having the user select the desired stream quality.

In comparison to the Isengard version of this addon, I've added a m3u8 parsing method to snnow.py, updated test.py to call the method, and updated default.py to display the different streams in a pop-up before playback.

TIA

### Checklist:


MartijnKaijser:
Please follow https://github.com/xbmc/repo-plugins/blob/master/CONTRIBUTING.md  on how to create PRs. Making one from Krypton branch itself is not recommended as you end up with the following.
If you are dropping support for anything older than Krypton please create a PR to mark it broken for those including version bump. Make sure the Krypton version is bumped enough to leave some version room between them.

In future consider switching to following language layout http://kodi.wiki/view/language_support#Layout_of_the_strings.po_file

PR #943, opened 2017-02-04
[plugin.video.redbull.tv] 1.2.2

keithah:
Marking the add-on as broken, since it throws an error when launched.


kib:
Looks good to me.

PR #950, opened 2017-02-06
[plugin.video.dmaxitalia] broken

Rechi:
Mark the add-on broken, as the website has changed and it doesn't work any longer.


enen92:
Thanks

PR #979, opened 2017-02-14
[plugin.video.revision3] broken

enen92:
Marking this plugin as broken

-Last update was 4 years ago
-I attempted to submit fixes for some broken sections (didn't receive feedback for almost 2 years)
-Author is away from our forums for 2 years

So, decided to do a full re-write and to be the new maintainer of the addon, since I use it a lot myself. The original author has been added as an author in the new addon.
My version will be submitted through another pull request.


ronie:
if you're going to keep the same addon id, i would first add your new version to the krypton repo before marking the old one as broken.

enen92:
Sure, just sent it 👍

PR #983, opened 2017-02-14
[plugin.video.viaplay] - deprecated

emilsvennesson:


enen92:
Version bump is needed as well. Make sure you choose a version lower than the one in the krypton Repo.

emilsvennesson:
@enen92 Are you sure? This is what the wiki says: The <broken> tag will mark the add-on as broken in the Kodi repo and provide the reason why. You don't need to do a version bump for this to work. However a bump is recommended as you could also add this to the changelog.

enen92:
If you dont bump the version it will only take effect for new addons instals. If you bump the version everyone using Jarvis will receive the notification and will know how to proceed/continue to receive support and updates. It is a good practice.

emilsvennesson:
@enen92 Alright, thanks. I screwed this branch up badly when I squashed my commits... Please merge #984 instead.

Sorry about that.

PR #1034, opened 2017-03-07
[plugin.video.latelelibre_fr] 1.0.2

jamontes:
### Description
Hi, this is an update on the Gotham branch for the plugin.video.latelelibre_fr

Changelog: v1.0.2 (2017.03.07)
- Sanitize thumbnail URLs (credits: andersonvom).
- Fix playable video URL pattern (credits: andersonvom).
- Undesired whitespace code cleaning (credits: andersonvom).
- Updated test case code (credits: andersonvom).
- Added support for preferred video quality option from add-on settings menu.
- Updated logging system.
- Improved video patterns.
- Improved special character translations from description fields.
- Code refactored.
 
Thanks a lot in advance and best regards,

jamontes.
### Checklist:

- [ X] My code follows the [add-on rules](http://kodi.wiki/view/Add-on_rules) and [piracy stance](http://kodi.wiki/view/Official:Forum_rules#Piracy_Policy) of this project. 
- [ X] I have read the [CONTRIBUTING](https://github.com/xbmc/repo-plugins/blob/master/CONTRIBUTING.md) document
- [ X] Each add-on submission should be a single commit with using the following style: [plugin.video.foo] v1.0.0

### DIFF
--- a/plugin.video.latelelibre_fr/resources/lib/lutil.py
+++ b/plugin.video.latelelibre_fr/resources/lib/lutil.py
@@ -1,74 +1,80 @@
 # _*_ coding: utf-8 _*_
 
 '''
-   lutil: library functions for XBMC video plugins.
-   Copyright (C) 2013 José Antonio Montes (jamontes)
+   lutil: library functions for KODI media add-ons.
+   Copyright (C) 2017 José Antonio Montes (jamontes)
 
    This program is free software: you can redistribute it and/or modify
    it under the terms of the GNU General Public License as published by
    the Free Software Foundation, either version 3 of the License, or
    (at your option) any later version.
-   
+
    This program is distributed in the hope that it will be useful,
    but WITHOUT ANY WARRANTY; without even the implied warranty of
    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
    GNU General Public License for more details.
-   
+
    You should have received a copy of the GNU General Public License
    along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
    Description:
    These funtions are called from the main plugin module, aimed to ease
    and simplify the plugin development process.
-   Release 0.1.7
+   Release 0.1.10
 '''
 
 # First of all We must import all the libraries used for plugin development.
 import re, urllib, urllib2
+from datetime import date
 
 debug_enable = False # The debug logs are disabled by default.
 
-def set_debug_mode(debug_flag):
-    """This function sets the debug_enable var to log everything if debug option is true."""
-    global debug_enable
-    debug_enable = debug_flag in ("true", True)
 
+def local_log(message):
+    """This function logs the messages into the main KODI log file. Called from the libraries module by other functions."""
 
-def log(message):
-    """This function logs the messages into the main XBMC log file. Called from main plugin module."""
     if debug_enable:
         print "%s" % message
 
 
-def _log(message):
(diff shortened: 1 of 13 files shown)


plugin.video.latelelibre_fr/resources/lib/lutil.py
```diff
@@ -1,74 +1,80 @@
...
-    global debug_enable
-    debug_enable = debug_flag in ("true", True)
 
+def local_log(message):
+    """This function logs the messages into the main KODI log file. Called from the libraries module by other functions."""
 
-def log(message):
-    """This function logs the messages into the main XBMC log file. Called from main plugin module."""
     if debug_enable:
         print "%s" % message
```
enen92: please replace this print statement with `xbmc.log` method with the log level set to `xbmc.LOGDEBUG`(see: http://mirrors.kodi.tv/docs/python-docs/13.0-gotham/xbmc.html).
Example: `xbmc.log(msg='This is a test string.', level=xbmc.LOGDEBUG));`.

You can also drop the if condition and the addon debug setting all together. It looks a bit more sane to enable debugging on a single location (on Kodi itself) than having debug settings hidden in the code or in the addon settings.

jamontes: Hi enen92,

The log system is already modified in this current PR to use `xbmc.log` function under Kodi environment on all cases (as far as the debug option is activated from the add-on settings menu), and `local_log` function (using the `print` statement) when the add-on API code (i.e. `ltl_api.py`) is run outside the Kodi environment (i.e only during API testing and development) if the local debug flag is activated. Whenever i mention API on this context i mean the API code to retrieve the media contents from the add-on's website (and not the Python Kodi API).

The `xbmc.log` function is called from `plugin.log` function located at `plugin.video.latelelibre_fr/resources/lib/plugin.py`.

The `plugin.log` function is passed from `default.py` as a parameter in `set_debug` function call to both `ltl_api.py` and `lutil.py` (`default.py` calls `ltl_api.set_debug` on line 48, and `ltl_api.set_debug` calls `lutil.set_debug_mode` on line 35 of `ltl_api.py` to pass to `lutil.py` the debug flag and log function used during Kodi runtime), so whenever Kodi runs the add-on, `plugin.log` (which calls `xbmc.log` for logging) is used in case of add-on settings debug option is activated. You can check it by running the add-on on your Kodi testing environment.

I know that print statements for logging purposes on add-ons under Kodi must be avoided (and on the other hand, since several Kodi releases, they are completely useless as they would print nothing on the Kodi log file), and that's the reason why i have modified the logging code to use always `xbmc.log` during debugging under Kodi runtime, and `print` during debugging the API standalone (i.e. running out of Kodi). I cannnot use `xbmc.log` out of Kodi runtime environment, and it's completely useless (and pointless) to call `print` inside of Kodi. This solution allows me to use the right logging function to debug the code on both scenarios.

On the other hand, I would prefer to keep the debug option on the add-on settings menu rather than use the global Kodi debug option, as this sort of debug is add-on specific, and activating the global Kodi debug just to debug the add-on could led to too much information to screen on the log file.

Please, consider my explanations and the reason why i have taken this solution to the debug log approach.

Thanks a lot in advance for your help and effort.

Best regards,

jamontes.

enen92: Oh okay, haven't noticed that :) Thanks for the explanation. Merged

PR #1057, opened 2017-03-18
[plugin.video.sgtv] 1.8.0

GadgetReactor:
### Description
Added Search function
Code cleanup

### Checklist:


enen92:
please change the PR title to "[plugin.video.sgtv] 1.8.0"

plugin.video.sgtv/default.py
```diff
@@ -108,18 +107,31 @@ def channelShows(channel, page=0):
 	if current_page < max_page:
 		page = str(current_page + 1)
 		addXBMCItem (__language__(31000), "", "?mode=loadChannel&channel="+channel+"&page="+page, True)
-
 	xbmc.executebuiltin("Container.SetViewMode(500)")
```
enen92: Please drop any calls to SetViewMode from the code. Forcing skin views are not allowed on the repository anymore I'm afraid. Skin view id's also depend on the skin the user is running. So please use xbmcplugin.setContent to define the correctly content type (tvshows for instance). Also please set the mediatype infolabel when building the ListItem object

plugin.video.sgtv/resources/search.xml
```diff
@@ -0,0 +1,2 @@
+(lp0
```
enen92: Not sure what is the search.xml. Does it really need to be shipped with the addon? This is probably best placed in the addon userdata

GadgetReactor: Thanks. Do you know where the addon userdata folder is? or how to reference it from the addon?

enen92: You can define it as `__profile__ = xbmc.translatePath(__addon__.getAddonInfo('profile')).decode('utf-8')` and then just write to a `search.xml` file in that folder. The file does not exist when you ship the addon but it will be created once the user starts to use the addon. Note that this change is not mandatory but it is generally a good principle. If you ship the search.xml within the addon, it will be re-written every update. So, taking into account this file is to store the users' search history, it will be lost every time you issue an update.

GadgetReactor:
Thanks - have made the changes, please help me review again.

enen92:
Thanks for adjusting (Y)
Merged

PR #1089, opened 2017-04-02
[plugin.video.rocketbeans] 1.0.0

Razzeee:
### Description

Basically wrapping the youtube plugin and adding some shedule reader via an api.

### Checklist:


Razzeee:
@enen92 in that case we need to update the readme https://github.com/xbmc/repo-plugins/blob/jarvis/README.md#status

enen92:
I'm guiding myself by the wiki: http://kodi.wiki/view/Submitting_Add-ons

Razzeee:
thought that would be the case, well one of them is not matching then
@MartijnKaijser what's correct?

enen92:
pinging @ronie and @phil65 as well

ronie:
wiki is always correct.

i wouldn't add this info to the readme, unless someone enjoys updating them for all repos with every release.

Razzeee:
Switched to krypton as a base

plugin.video.rocketbeans/addon.xml
```diff
@@ -0,0 +1,27 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<addon id="plugin.video.rocketbeans" version="1.0.0" name="Rocket Beans TV" provider-name="Razzeee">
+    <requires>
+        <import addon="xbmc.python" version="2.14.0" />
```
enen92: bump

plugin.video.rocketbeans/addon.xml
```diff
@@ -0,0 +1,27 @@
...
+        <import addon="plugin.video.youtube" version="5.2.1" />
+        <import addon="script.module.routing" version="0.2.0"/>
+    </requires>
+    <extension point="xbmc.python.pluginsource" library="addon.py">
+        <provides>video</provides>
+    </extension>
+    <extension point="xbmc.addon.metadata">
+        <summary lang="de_de">Rocket Beans TV | 24/7 Entertainment</summary>
```
enen92: I think you are required to provide the same tags in en_GB as well (for the website and addon browser)

plugin.video.rocketbeans/addon.xml
Razzeee: Yeah, that sliped between the cracks now

Razzeee:
Updated

PR #1203, opened 2017-05-10
[plugin.video.mediathek] 0.8.3

raptor2101:
### Description
Fix to get arte working again
### Checklist:

### DIFF
--- a/plugin.video.mediathek/mediathek/arte.py
+++ b/plugin.video.mediathek/mediathek/arte.py
@@ -19,6 +19,7 @@ import re, traceback,json;
 from mediathek import *;
 from bs4 import BeautifulSoup;
 
+
 regex_dateString = re.compile("\\d{1,2} ((\\w{3})|(\\d{2})) \\d{4}");
 month_replacements = {
     "Jan":"01",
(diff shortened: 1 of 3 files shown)


plugin.video.mediathek/mediathek/arte.py
```diff
@@ -195,15 +163,45 @@ def buildVideoEntry(self, jsonObject):
...
-    duration = jsonObject["duration"]
-    self.gui.buildVideoLink(DisplayObject(title,subTitle,picture["url"],detail,link,"JsonLink", pubDate,duration),self,0);
+    if("teaser" in jsonObject):
+      detail = unicode(jsonObject["teaser"]);
+    else:
+      detail = "";
+
+    pictures = None;
+    pictureUrl = None;
+    if("thumbnails" in jsonObject):
```
learningit: Just a comment here. You could use pictures = jsonObject.get("thumbnails") in place of the if statement. The second parameter of the .get() is the value to return if the index isn't found, which defaults to None.

raptor2101: thanks for the hint

PR #1205, opened 2017-05-14
[plugin.onedrive] 1.2.0

cguZZman:
v1.2.0 released:
- Download subtitle fixed
- Now supporting business accounts
- xbmcvfs is used for disk operations

### DIFF
--- a/plugin.onedrive/addon.py
+++ b/plugin.onedrive/addon.py
@@ -23,8 +23,6 @@
 import ConfigParser
 import json
 import os
-import shutil
-import stat
 import sys
 import threading
 import time
@@ -34,12 +32,12 @@ import urllib2
 import urlparse
 
 from resources.lib.api import utils
-from resources.lib.api.onedrive import OneDrive, OneDriveException, \
-    AccountNotFoundException
+from resources.lib.api.onedrive import OneDrive, OneDriveException, AccountNotFoundException
 import xbmc
 import xbmcaddon
 import xbmcgui
 import xbmcplugin
+import xbmcvfs
 
 
 base_url = sys.argv[0]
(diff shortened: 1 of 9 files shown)


plugin.onedrive/addon.py
```diff
@@ -307,29 +300,22 @@ def export_folder(name, item_id, driveid, destination_folder, base_folder, direc
...
-                fo.write(url)
+            f = xbmcvfs.File(parent_folder + name + '.strm', 'w')
+            f.write(url)
+            f.close()
         onedrive.exporting_count += 1
         p = int(onedrive.exporting_count/float(onedrive.exporting_target)*100)
         if onedrive.exporting_percent < p:
             onedrive.exporting_percent = p
-        file_path = os.path.join(parent_folder, name)
+        file_path = parent_folder + name
```
enen92: please use os.path.join here. It is a bad practice to concatenate strings for folder paths

plugin.onedrive/addon.xml
```diff
@@ -1,25 +1,25 @@
...
 		<provides>image audio video</provides>
 	</extension>
+	<extension point="xbmc.service" library="service.py" start="login" />
 	<extension point="xbmc.addon.metadata">
 		<platform>all</platform>
 		<summary lang="en">Microsoft OneDrive for KODI</summary>
 		<description lang="en">
 Play all your media from Microsoft OneDrive including Videos, Music and Pictures.
- - Unlimited number of accounts.
+ - Unlimited number of accounts. Personal or Business.
```
enen92: Not sure if it applies but probably a good idea do use [CR] for line breaks. Not a blocker though.

PR #1221, opened 2017-05-24
[plugin.video.iplayerwww] 3.0.7

CaptainTK:
### Checklist:

Additional information :

### DIFF
--- a/plugin.video.iplayerwww/addon.xml
+++ b/plugin.video.iplayerwww/addon.xml
@@ -1,9 +1,8 @@
 <?xml version="1.0" encoding="UTF-8" standalone="yes"?>
-<addon id="plugin.video.iplayerwww" name="iPlayer WWW" version="3.0.6" provider-name="CaptainT, Cas, ihurst, primaeval">
+<addon id="plugin.video.iplayerwww" name="iPlayer WWW" version="3.0.7" provider-name="CaptainT, Cas, ihurst, primaeval">
   <requires>
     <import addon="xbmc.python" version="2.25.0"/>
     <import addon="script.module.requests" version="2.7.0"/>
-    <import addon="inputstream.adaptive" version="1.0.6"/>
   </requires>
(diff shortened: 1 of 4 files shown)


MartijnKaijser:
Did you take in account that inpustream is disabled by default and tested if it actually works like you intended?

CaptainTK:
Yes, I did. Switched DASH on, tried Live with and without Autoplay as well as Catchup with and without Autoplay. In between I switched DASH back on and it fell back to HLS as expected. I tested with 17.1 on Linux and simply disabled InputStream Adaptive for the tests.

MartijnKaijser:
No i mean if auto enables the inpustream addon cause that's what the dependency did. Now you rely on users to enable it theirselves

CaptainTK:
Ah, sorry, didn't get that before.

No, this code does not automatically enable inputstream. It simply avoids errors if it is not available. Users now need to activate inputstream and then switch stream protocol to DASH, otherwise it will stick to HLS.

The dependency worked the same way, users had to enable inputstream, otherwise the add-on failed to install. If this was supposed to work automatically as you indicate, then this automatic procedure is broken. It always seemed to me that inputsream, being a binary add-on, was not installed the same way as non-binary add-ons like requests.

HLS is the default stream protocol (again) now. Apart from the issues with inputstream, this is also because there are some programmes that are not available as DASH streams and currently they simply won't play. I tried implementing an automatic fallback, but failed before. Will need to revisit this.

Frankly speaking, I am a little tired by the amount of issues inputstream caused. I was very enthusiastic when Krypton brought support for DASH, but the whole concept how inputstream is invoked seems flawed to me.

Also, I would like to switch to a different mediaselector, but this is blocked by a bug in inputstream. I opened an issue for it, but nothing happened so far.

MartijnKaijser:
Inputstream is still in very early stage so no need to shoot it down

PR #1223, opened 2017-05-26
[plugin.video.tyt] 2.1.2

lottaboost:
Sorry for this update following so closely to the last update.  I will address this recent issue with content provider changing URLS frequently on my next update, which will be in a week or two, with no more updates till then.

Brian


MartijnKaijser:
@lottaboost could you please fix this?
PROBLEM: Not found LICENSE\.txt|LICENSE\.md|LICENSE in folder /home/travis/build/xbmc/repo-plugins/plugin.video.tyt

lottaboost:
There is a license.txt file, does it need to be in all caps?


Brian


________________________________

Razzeee:
There is a bug in the checker script, it only looks for all caps, but I already fixed it. So don't change anything. Thanks for replying this quick :)

PR #1278, opened 2017-06-24
[plugin.video.embycon] 1.3.15

faush01:
### Description

Addon to display and play back media libraries from the Emby Server.

### Checklist:

Additional information :


faush01:
Is this the correct place to submit addons?
Been a week or so since submission, just wondering if I have submit this correctly.

If you are looking for more info about the addon you can check the emby server forums:
https://emby.media/community/index.php?/topic/46651-embycon/

Tadly:
It's the right place and I started reviewing it too.
This just takes time with bigger add-ons and I don't have as much time to spend on reviewing lately.

plugin.video.embycon/resources/lib/kodi_utils.py
```diff
@@ -0,0 +1,64 @@
...
+    def clearProperty(self, key):
+        key = self.id_string % key
+        # log.debug('HomeWindow: clearProperty |%s|' % key)
+        self.window.clearProperty(key)
+
+
+def addMenuDirectoryItem(label, path, folder=True, thumbnail=None):
+    li = xbmcgui.ListItem(label, path=path)
+    if thumbnail is None:
+        thumbnail = "special://home/addons/plugin.video.embycon/icon.png"
```
Tadly:
```
addon = xbmcaddon.Addon()  # have that global like the logger maybe
addon.getAddonInfo('icon') # <- this is the preferred way of getting the addons icon (or other addon related infos)
```

plugin.video.embycon/resources/lib/simple_logging.py
```diff
@@ -0,0 +1,46 @@
...
+    def getLevel(self):
+        return self.level
+
+    def __str__(self):
+        return "LogLevel: " + str(self.level)
+
+    def error(self, msg):
+        if (self.level >= 0):
+            try:
+                xbmc.log(self.format(msg, "ERROR"), level=xbmc.LOGNOTICE)
```
Tadly: why not `level=xbmc.LOGERROR`?

plugin.video.embycon/resources/lib/simple_logging.py
```diff
@@ -0,0 +1,46 @@
...
+            except UnicodeEncodeError:
+                xbmc.log(self.format(msg, "ERROR").encode('utf-8'), level=xbmc.LOGNOTICE)
+
+    def info(self, msg):
+        if (self.level >= 1):
+            try:
+                xbmc.log(self.format(msg, "INFO"), level=xbmc.LOGNOTICE)
```
Tadly: why not `level=xbmc.LOGINFO`

plugin.video.embycon/resources/lib/simple_logging.py
```diff
@@ -0,0 +1,46 @@
...
+            except UnicodeEncodeError:
+                xbmc.log(self.format(msg, "INFO").encode('utf-8'), level=xbmc.LOGNOTICE)
+
+    def debug(self, msg):
+        if (self.level >= 2):
+            try:
+                xbmc.log(self.format(msg, "DEBUG"), level=xbmc.LOGNOTICE)
```
Tadly: this one make sure to set to `level=xbmc.LOGDEBUG` so kodis logfile doesn't get polluted with debug messages even iff debugging is disabled.

plugin.video.embycon/resources/lib/views.py
```diff
@@ -0,0 +1,195 @@
...
+        jsonData = dataFile.read()
+        dataFile.close()
+        defaultViewData = json.loads(jsonData)
+        log.info("Loaded skin views: " + str(defaultViewData))
+        skin_used = xbmc.getSkinDir()
+        log.info("Current skin: " + skin_used)
+        skin_views = defaultViewData.get(skin_used, None)
+        log.info("Current skin views: " + str(skin_views))
+        if skin_views is None:
+            xbmcgui.Dialog().notification(__addon__.getAddonInfo('name'), i18n('skin_not_supported') % skin_used, icon='special://home/addons/plugin.video.embycon/icon.png')
```
Tadly: for the icon use `__addon__.getAddonInfo('icon')` instead

plugin.video.embycon/service.py
```diff
@@ -0,0 +1,371 @@
...
+
+ga = GoogleAnalytics()
+try:
+    ga.sendEventData("Version", "OS", platform.platform())
+    ga.sendEventData("Version", "Python", platform.python_version())
+except Exception as error:
+    log.error("Exception in sending client meta info: " + str(error))
+
+try:
+    while not xbmc.abortRequested:
```
Tadly:
```python
monitor = xbmc.Monitor()
while not monitor.abortRequested():
    ...
```
further down, replace the `xbmc.sleep(1000)` with `monitor.waitForAbort(1)`

plugin.video.embycon/service.py
```diff
@@ -0,0 +1,371 @@
...
+                log.error("Exception in Playback Monitor : " + str(error))
+
+        else:
+            play_data = home_window.getProperty("play_item_message")
+            if play_data:
+                home_window.clearProperty("play_item_message")
+                play_info = json.loads(play_data)
+                playFile(play_info)
+
+        xbmc.sleep(1000)
```
Tadly: just to make sure, that's what I meant...
`monitor.waitForAbort(1)`

plugin.video.embycon/resources/lib/views.py
```diff
@@ -0,0 +1,195 @@
...
+        log.info("WINDOW INITIALISED")
+
+    def onInit(self):
+        self.action_exitkeys_id = [10, 13]
+
+        # load skin views               
+        addonPath = __addon__.getAddonInfo('path')
+        skin_view_file = os.path.join(addonPath, "resources", "data", "skin_views.json")
+        log.info("Loading skin views form: " + skin_view_file)
+        dataFile = open(skin_view_file, 'r')
```
Tadly: This ensures proper file handling (closing in case of an error etc.)
```python
with open(skin_view_file, 'r') as dataFile:
    jsonData = dataFile.read()
defaultViewData = json.loads(jsonData)
```

Tadly:
A few more things.

1. As of our [addon-rules](http://kodi.wiki/view/Add-on_structure#icon.png), your icon should not have transparency.

2. The whole `GoogleAnalytics` thing. As far as I can tell there's no way of opting out and personally I'd prefer it to be opt-in all together. I asked the team what they think and will come back to you on that. But having a way of turning it on/off in the settings **will** be the minimum.

3. Can you please explain the use of `default_views.json` to me? Reason being that we don't like addons changing the current view.
Instead, listitems should be prepared properly (setting media-types etc.) and skins should work off of those information's. With the limited time I have I couldn't quite figure out what exactly it is you're are doing so you shall explain it :)

plugin.video.embycon/resources/lib/simple_logging.py
faush01: the problem with this is if you want to show addon debugging log info you need to enable debugging logging in two places, the addon and also Kodi, this also prints out all the Kodi debug info which causes logs to be huge and hard to manage for submission.
If this is a problem I guess I can do this but it will make my life harder.

Tadly: I'm not sure if I can follow. Why in two places?
You, as an addon, are supposed to do simple logging with a specific level. Kodi by default logs everything down to... erm.. info I believe.
If a user has an issue and you require debug logs than he has to enable debug-logging within kodi, reproduce and send that log to you.

Yes, these logs will be big **but** you have formatted logs including your add-on name as a 
prefix hence you have an easy time searching for your lines only.

faush01:
1 - I will update the icon

2 - You can opt out in the settings, under advanced "Log error and Anonymous metrics" perhaps this need qualification "Remotely Log Errors and Anonymous Metrics"
Adding remotely should clear that up. 

3 - By default no view is set, it leaves it up to the Skin and Kodi.
However this is not ideal as even when you set the correct media type in the addon for the list items and want to set a new view for say all your seasons or episodes you need to go in and set the view you want in each and every path, i.e. go into tv show X and set the seasons to wide list, set the episodes for season one to wall, go into season 2 and set the view to wall etc, each and every location needs to have the view set.
This option lets you set the view type for each media Type and it is applied if you view media of that type, it does this on a Skin by Skin basis, that is the view IDs need to be available in the defined skin views file and then you need to go inot the addon menu and select Set Default Views to set the views you want for each media type for your selected skin.

Tadly:
Regarding Analytics...
Just got response form a member and he pointed me to our [Addon-Rules](http://kodi.wiki/view/Add-on_rules#Requirements_for_scripts_and_plugins) where it says:
> directly using analytics (Google Analytics for example) from within add-ons is not allowed. This should be handled server side by using user-agent.

So for use to be able to accept this add-on the whole analytics stuff needs to be removed

plugin.video.embycon/resources/lib/simple_logging.py
faush01: Yep that is my point I guess, the addon currently has a log level setting, so if I need a user to submit a log I tell them to set the addon logging to debug and reproduce the issue and submit the log.
The model you are referring to is that the addon would not have any log level setting at all. It would just log with the desired log level and set the correct KODI log level in the xbmc.log call
This approach I am using is probably left over from Kodi 14 or even 13, this addon has been alive for a long time.
If I really need to change this I can I guess but as I said it will make my life harder, dealing with bigger log files was never fun and in fact when I see a user log submission with both the addon and kodi debug logging turned on I usually get them to re run with just the addon logging set to debug as trying to follow the large Kodi Debug log file is a pain.

faush01:
Can you explain what this means
 > This should be handled server side by using user-agent.

Tadly:
Regarding point 3. I know problems (Been there myself with a google music plugin) however we do not accept this as of our [Addon-Rules](http://kodi.wiki/view/Add-on_rules#Requirements_for_scripts_and_plugins):
> Plugins should not force certain skin viewtypes on their own. They should set the correct media content from which the correct skin view type is automatically provided.

As a personal note:
We (kodi) try to constantly improve the core and since krypton I don't have to set any views for my google music add-on anymore as the defaults (due to properly set mediatypes etc.) are just fine.

Tadly:
Can't really explain it as I didn't write that rule myself but I'd assume this means that in your case, emby itself should do analytics, not the kodi add-on.

Same would apply to other add-ons like plex, gmusic etc.

faush01:
Ok, so to move forward I would need to:

 - remove all remote error and metrics logging
 - revamp the internal addon/kodi logging
 - remove the default view type setting

Anything else?

the first two are internal and only effect devs, the last one I would run past the user base on the emby forums to see if it was a deal breaker or not.

plugin.video.embycon/resources/lib/simple_logging.py
Tadly: I understand how frustrating it is having to deal with big logs, I really do but it's the preferred way non the less.

To maybe ease future log-reading here's what I'd do in your case.

**On linux:** (I assume you're a windows user but just in case you're not)
```bash
# Will print to the terminal
cat kodi.log | grep embycon

# Will save the output to a new file
cat kodi.log | grep embycon > kodi.new.log
```
`grep` is a tool that. Sadly windows does (still) not come with any similar tool AFAIK

**On windows:**
Quick googeling suggests that this is a great alternative for windows.
http://stefanstools.sourceforge.net/grepWin.html

Tadly:
Sounds good, yes.

Regarding the last point. Maybe turn the custom-view types of and test your add-on with estuary.
The default views estuary picks should be just fine (unless you forgot to set a mediatype etc. which you should double/tribble check at this point).

If estuary looks good but other skins don't it suggests that (whatever skin it may be) needs some work to properly support all mediatypes.

faush01:
It still only sets a view for a particular path, from my understanding the view is stored in the DB for each path. This means that for each and every season or episode list I wanted to set a view for I would have to go in and set that view.
As I pointed out above this would need to happen for each path in the the library.

I am sure the addon is setting the media type correctly, well actually you can tell me you can see the code in the pull request, look for the getContent() call in functions.py

How should it work, am I seeing the correct functionality of the default Kodi or is something wrong?

Tadly:
I know the problem that arises (as mentioned before, I hit it myself) and you are right in assuming that the user would than have to change the view for every single element (as the path changes due to ids etc.)

faush01:
So if this is a know issue why are addons not allowed to set the view?
What was the original reasoning behind this? who came up with this?

Tadly:
It's not an issue, it's just a limitation which has been improved upon over the years and will be further improved in coming years.
Forcing a viewtype is considered to be hacky'isch as kodi will load a specific view (depending on the content) first and only after you're able to force-overwrite it again.
That's a bad sequence of events (as you can imagine).

I bet there're other reason too though...

faush01:
ok point taken. I will work on this over the weekend.

Tadly:
I'm still discussion internally but it's not looking that great, sorry.
The gist of it is that viewtypes and their codes are skin-dependent hence it's bad for add-ons to change it.

faush01:
I completely understand you dont want display decisions being made by addons, that really should be skins that do that.

Perhaps there should be a built in way for Kodi to apply a Skin View for all content of that type.
i.e.
You navigate to episodes, there is an option (built in function) that you can select that for all content of the type you are currently viewing you use the current view ID from that point on.
This would be very similar to the current approach where the Skin sets the default view for best user experience but if the user wants they can say use this view type for all media of this type from now on.

faush01:
I some how screwed up this pull request. I have created another one

https://github.com/xbmc/repo-plugins/pull/1309

All the requested changes should now be in the new pull request.
This one can be closed.

faush01:
I am going to close this one as it us redundant now, see https://github.com/xbmc/repo-plugins/pull/1309 for the new pull request with all the changes in it.

PR #1368, opened 2017-08-10
[plugin.audio.mixcloud] v2.2.0

jackyNIX:
- added new resolver to fix broken mixcloud links
- modified french localisation

### DIFF
--- a/plugin.audio.mixcloud/addon.xml
+++ b/plugin.audio.mixcloud/addon.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8" standalone="yes"?>
 <addon id="plugin.audio.mixcloud"
        name="MixCloud"
-       version="2.1.3"
+       version="2.2.0"
        provider-name="jackyNIX">
   <requires>
     <import addon="xbmc.python" version="2.1.0"/>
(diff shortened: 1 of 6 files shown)


plugin.audio.mixcloud/addon.xml
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8" standalone="yes"?>
 <addon id="plugin.audio.mixcloud"
        name="MixCloud"
-       version="2.1.3"
+       version="2.2.0"
        provider-name="jackyNIX">
   <requires>
     <import addon="xbmc.python" version="2.1.0"/>
```
enen92: while you are on it can you please bump the xbmc.python version to 2.14.0 so it matches the gotham python api version?

plugin.audio.mixcloud/resources/language/French/strings.xml
```diff
@@ -3,44 +3,44 @@
 	<!-- common -->
```
enen92: Something you might want to consider for the future is to adapt the strings.xml to the new strings.po language layout: http://kodi.wiki/view/Language_support

enen92:
thanks

plugin.audio.mixcloud/addon.xml
jackyNIX: done.
but the rebase thing to squash the commit is not so clear to me...

enen92: We can squash commits on merging, no problem ;)

PR #1389, opened 2017-08-31
[plugin.video.manoto] 2.0.1

bbk79:
### Description
Updated the stream URL
### Checklist:

Additional information :

### DIFF
--- a/plugin.video.manoto/default.py
+++ b/plugin.video.manoto/default.py
@@ -7,5 +7,5 @@ li = xbmcgui.ListItem ('manoto')
 li.setArt({ 'thumb': icon })
 li.setInfo('video', {'plot': 'Live Stream'})
 
-url = 'http://manotolive-i.akamaihd.net/hls/live/251779/livexhq/master_Layer1.m3u8'
+url = 'https://d1v7e0o3q8z5j0.cloudfront.net/Live.m3u8'
 xbmc.Player().play(url, li, False)
--- a/plugin.video.manoto/addon.xml
+++ b/plugin.video.manoto/addon.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8" standalone="yes"?>
 <addon id="plugin.video.manoto"
        name="Manoto TV"
-       version="2.0.0"
+       version="2.0.1"
        provider-name="babak">
   <requires>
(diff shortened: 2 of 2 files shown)


plugin.video.manoto/default.py
```diff
@@ -7,5 +7,5 @@
 li.setArt({ 'thumb': icon })
 li.setInfo('video', {'plot': 'Live Stream'})
 
-url = 'http://manotolive-i.akamaihd.net/hls/live/251779/livexhq/master_Layer1.m3u8'
+url = 'https://d1v7e0o3q8z5j0.cloudfront.net/Live.m3u8'
```
enen92: Probably a better idea to write a resolver for the livestream otherwise you need to hardcode the stream link all the time with every update

Lunatixz: @enen92 I don't mean to pry but I was following this commit out of curiosity. Is the use of an arbitrary live link allowed? As is there is no way to verify its origins... I only submit plugins with live feeds I can resolve directly from the website... I have a few other plugins that I know the links are valid from the source but I am unable to resolve directly from the website. Would it be okay to submit those plugins to the repo?

bbk79:
FWIW there used to be a resolver in pre 2.0.0 versions of the addon as the URL was dynamic before then, back in February they started using a static link and I removed all the resolver logic and now that static link has changed as they have apparently switched from Akamai to Cloudfront.

Nevertheless, I agree, using a resolver is a better idea.

plugin.video.manoto/default.py
bbk79: FWIW there used to be a resolver before version 2.0.0 of the add-on and the link was a dynamic one. I removed the resolver when they switched to a static address back in February and now they have apparently switched from Akamai to Cloudfront.

Nevertheless, I agree, using a resolver is a better idea and I'll go back to that in next update.

enen92: @Lunatixz I checked myself if it was the link provided on the website (and also checked the previous one on the previous update). This is one of the reasons I said a resolver is a better option - at least easier to check.
I guess you're free to submit them to the repo (as long as the source can be verified even if only manually like this one).

Lunatixz: @enen92 Okay, thanks for the info =)

PR #1450, opened 2017-10-09
 [plugin.video.embycon] 1.3.40

faush01:
### Description

Latest version of EmbyCon
 - New action menu for widgets
 - New Widget actions for skins
 - Many fixes
### Checklist:

- [ ] My code follows the [add-on rules](http://kodi.wiki/view/Add-on_rules) and [piracy stance](http://kodi.wiki/view/Official:Forum_rules#Piracy_Policy) of this project. 
- [ ] I have read the [CONTRIBUTING](https://github.com/xbmc/repo-plugins/blob/master/CONTRIBUTING.md) document
- [ ] Each add-on submission should be a single commit with using the following style: [plugin.video.foo] v1.0.0

Additional information :


plugin.video.embycon/addon.xml
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8" standalone="yes"?>
 <addon  id="plugin.video.embycon" 
         name="EmbyCon"
-        version="1.3.24"
+        version="1.3.40"
         provider-name="Team B">
   <requires>
     <import addon="xbmc.python" version="2.24.0"/>
```
Lunatixz: Please change to 2.25.0

faush01:
From memory this stops the addon being comparable with Kodi 16 though.

Lunatixz:
Correct, however you are committing this PR to Krypton. The Jarvis branch is closed.

faush01:
People on Kodi 16 are still using the add-on and update from the test repo I have set up, changing this would break that option for them. Alternatively I would need to create a new add-on that is Kodi 16 comparable as I can't have two add-ons in the same repo with the same id, I don't this that is manageable.

Lunatixz:
I understand, however this commit was originally made to Krypton... The Jarvis branch is closed. I would suggest you create a new branch in your personal repo for "jarvis" builds and continue to update the kodi repository with Krypton updates. Jarvis users with your repository will receive updates, while their systems ignore Krypton updates via the Official repository.

faush01:
That's the problem, maintaining both, keeping them updated. These is no reason for it, the addon works on both but now I need a seperate branch to support both.

Lunatixz:
http://kodi.wiki/view/Submitting_Add-ons#Allowed_submissions

faush01:
updated, I will work out what I am going to do with K16 some other time.

learningit:
@faush01 I don't think that made a difference anyway. A Jarvis user would not see your add-on in the kodi.org repo because it is in the Krypton branch, regardless the xbmc.python being set for Jarvis.

MartijnKaijser:
correct. What's in Krypton will only be in Krypton and up. V16 is dead to us so we don't want new addons for it. I'm not opposed to having a lower xbmc.python version in case they want to install it manually

faush01:
Thanks for the clarification.
That is exactly what I wanted to do, just have it so users could install the addon manually if they wanted to or needed to for Kodi 16. I never intended to actually submit the addon to the old Kodi 16 addon branch.

faush01:
So can we merge this change then?

faush01:
Can someone please explain what I need to do here to have this merged.

I just need clear directions on what is holding up the process so I can fix it.

Lunatixz:
@faush01 I believe you need a second review; larger programs such as your's requires more time. From what I can tell you are good to go after a second review :).

Lunatixz:
@enen92 @Razzeee LGTM

PR #1470, opened 2017-10-22
[plugin.video.dailymotion_com] 2.3.3

Gujal00:
### Description

The current maintainer L0RE has moved on to develop a different addon and this codebase has been left in limbo. As per L0RE, he is happy to handover maintenance of this addon to me.
https://forum.kodi.tv/showthread.php?tid=142626&pid=2656751#pid2656751
The changes include the following:
Make videos playable in Krypton
Cleanup LiveTV section to show only active streams.
Change artwork to reflect current Dailymotion branding
Remove all deprecated options

### Checklist:

- [X] My code follows the [add-on rules](http://kodi.wiki/view/Add-on_rules) and [piracy stance](http://kodi.wiki/view/Official:Forum_rules#Piracy_Policy) of this project. 
- [X] I have read the [CONTRIBUTING](https://github.com/xbmc/repo-plugins/blob/master/CONTRIBUTING.md) document
- [X] Each add-on submission should be a single commit with using the following style: [plugin.video.foo] v1.0.0

Additional information :
- Submitting your add-on to this specific branch makes it available to any Kodi version equal or higher than the branch name with the applicable Kodi dependencies limits.
- [add-on development](http://kodi.wiki/view/Add-on_development) wiki page.
- Kodi [pydocs](http://kodi.wiki/view/PyDocs) provide information about the Python API
- [PEP8](https://www.python.org/dev/peps/pep-0008/) codingstyle which is considered best practice but not mandatory.
- This add-on repository has automated code guideline check which could help you improve your coding. You can find the results of these check at [Codacy](https://www.codacy.com/app/Kodi/repo-plugins/dashboard). You can create your own account as well to continuously monitor your python coding before submitting to repo.
- Development questions can be asked in the [add-on development](http://forum.kodi.tv/forumdisplay.php?fid=26) section on the Kodi forum.

PR #1472, opened 2017-10-22
Revert "[plugin.video.dailymotion_com] 2.3.3"

enen92:
Reverts xbmc/repo-plugins#1470

@Gujal00 please fix the issues mentioned in the original pull request: https://github.com/xbmc/repo-plugins/pull/1470 pointed by @Razzeee


Rechi:
@enen92 new version is already on the server revering it this way doesn't work as you expect.

Razzeee:
@Rechi 
well it kinda does, tomorrow http://mirror.de.leaseweb.net/xbmc/addons/jarvis/addons.xml will be updated, and it will get the newest commits from here, so it shouldn't include the reverted part.

Rechi:
no it won't, as you haven't increased the addon version

PR #1492, opened 2017-11-04
[plugin.video.myvevo] v3.0.8

idleloop-github:
### Description

version field in json answer doesn't seem valuable (there are values of 2, 3 or 4 at least), and in fact restricting it hides legit videos from loading.

### Checklist:

Additional information :

### DIFF
--- a/plugin.video.myvevo/README.txt
+++ b/plugin.video.myvevo/README.txt
@@ -4,6 +4,7 @@ plugin.video.myvevo
 
 Kodi Addon for VEVO
 
+Version 3.0.8 another m3u8 version fix
 Version 3.0.7 m3u8 version fix
 Version 3.0.6 auth fix
 Version 3.0.5 fix playlists, remove live channels, fix navigation
--- a/plugin.video.myvevo/addon.xml
+++ b/plugin.video.myvevo/addon.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8" standalone="yes"?>
 <addon id="plugin.video.myvevo"
        name="VEVO"
-       version="3.0.7"
+       version="3.0.8"
        provider-name="t1m">
   <requires>
     <import addon="xbmc.python" version="2.24.0"/>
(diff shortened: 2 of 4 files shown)


learningit:
It's not as simple as that. the version indicates what the video format is, or at least it used to.
Earlier versions of Kodi are unable to play several of those formats which is why it was being checked. it's possible that VEVO is no longer supporting stuff like DASH, but last I looked it was.

Even then, I believe there are a couple of formats that Kodi doesn't play (.wvm, etc) or are just audio. 
So possibly if the change is made to a Krypton or higher version it would be OK, but not for earlier versions. I'll keep this in mind though.

idleloop-github:
Ok! it resolved loading some videos for me, but your criterion is wiser.
Please, fell free to close this push request.
Thanks for your time!

idleloop-github:
needs more coding and testing

learningit:
I need to have a through look at tgis. VEVO is always in a state of flux. But I think we need to filter out some formats which we know won't play in earlie e Jodi versions. It is usually with older videos, so it's not always obvious when you test.

learningit:
I'm going to repoen this to remind me to look into the versioning. I most likely will need to create a Krypton and above version anyway and I think this maybe valid. In a recent change, VEVO has decided to return between 4 to 8 alternate urls for each bitrate in index.m3u8. This cause the auto detect for best stream in the player to go a bit crazy and time out on lower powered platforms. I may need to pull the index.m3u8 apart before passing it to the player to fix this.  Pre-Jarvis versions don't have the issue, or it's less apparent anyway.

idleloop-github:
Yeah, the m3u8 is full of resolution options and three backup urls for each of them. I also had to treat a (less complicated) m3u8 before passing it to the player [in another plugin](https://github.com/idleloop-github/xbmc-earthcam/blob/303047e49cd7ae080422bd9acaa676f8c34054f1/channel.py#L364).
It would be great if resolutions could be selected in addon's options. Something like SD, 720 & 1080. Just a suggestion ;-)

learningit:
Thanks for the time you spent on this. I've updated the add-on in the repo. I ended up using inputstream.adaptive to solve the backup url issue.

PR #1508, opened 2017-11-12
[plugin.googledrive] 1.0.0

cguZZman:
### Description
* New Google Drive addon.
* **Dependency on the new clouddrive module. PR: xbmc/repo-scripts#659**

### Checklist:


Lunatixz:
LGTM

plugin.googledrive/addon.py
```diff
@@ -0,0 +1,264 @@
+#-------------------------------------------------------------------------------
```
enen92: this is not mandatory at all but please consider to reduce the size of the entrypoint python file in the future (which is referenced in the addon.xml) since it is the only file that is not compiled by the python interpreter. It should be as small as possible and just serve as an entrypoint to the main logic. In other words, don't place all of your main code on the entrypoint python file.
You can see how it is done with the `generator-kodi-addon` : https://github.com/xbmc/generator-kodi-addon/blob/master/generators/app/templates/main.py

cguZZman: Got it. Will review the example and make the changes in a future.

PR #1562, opened 2017-12-20
[plugin.video.media-ccc-de] v0.1.1

TobiX:
### Description
- Supports multiple translated streams (if available)
### Checklist:

### DIFF
--- a/plugin.video.media-ccc-de/addon.xml
+++ b/plugin.video.media-ccc-de/addon.xml
@@ -1,8 +1,7 @@
 <?xml version="1.0" encoding="UTF-8" standalone="yes"?>
-<addon id="plugin.video.media-ccc-de" name="media.ccc.de" version="0.1.0" provider-name="Tobias Gruetzmacher">
+<addon id="plugin.video.media-ccc-de" name="media.ccc.de" version="0.1.1" provider-name="Tobias Gruetzmacher">
   <requires>
     <import addon="xbmc.python" version="2.24.0"/>
-    <import addon="script.module.simplejson" version="3.3.0"/>
     <import addon="script.module.requests" version="2.3.0"/>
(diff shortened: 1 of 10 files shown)


plugin.video.media-ccc-de/addon.xml
```diff
@@ -1,8 +1,7 @@
 <?xml version="1.0" encoding="UTF-8" standalone="yes"?>
-<addon id="plugin.video.media-ccc-de" name="media.ccc.de" version="0.1.0" provider-name="Tobias Gruetzmacher">
+<addon id="plugin.video.media-ccc-de" name="media.ccc.de" version="0.1.1" provider-name="Tobias Gruetzmacher">
   <requires>
     <import addon="xbmc.python" version="2.24.0"/>
-    <import addon="script.module.simplejson" version="3.3.0"/>
```
enen92: why did you remove simplejson? note that your addon is in the jarvis branch which means you cannot effectively rely on the json module in every kodi platform. Or does it only affect unit tests?

MartijnKaijser: you can certainly rely on json as it was in python2.5 and up. It just might be a bit slower

TobiX: I don't think the performance difference between simplejson and json is such a big deal and shouldn't everyone be on Python 2.7 nowadays (yeah, I know: hopes and dreams ;))?

Anyways, I was only using simplejson for tests and I rely on whatever requests is using (which might be simplejson AFAICS, but it also runs without) when parsing the "real" data.

Related: Are there statistics which Kodi versions are still in use? (e.g. addon repository meta data downloads split by user agent?)

MartijnKaijser: 3/4 is Krypton or up

PR #1572, opened 2017-12-27
[plugin.video.viaplay] 2.0.1

emilsvennesson:
### Description
+ New dependency: script.module.inputstreamhelper
+ Switch to registration code activation
+ Finally replace HLS with MPEG-DASH (Widevine DRM is now required)
+ Channels support (previously Viasat TV To Go)
+ Implement dynamically acquired 'theme pages' in addition to browsing through categories
+ Starred, Watched and Purchased pages now work
+ Open add-on settings on first run
+ Major rewrite and cleanup of the code
+ Big improvements all-around and a massively enhanced experience

### Checklist:

Additional information :


plugin.video.viaplay/resources/lib/viaplay.py
```diff
@@ -0,0 +1,347 @@
...
+        def __init__(self, value):
+            self.value = value
+
+        def __str__(self):
+            return repr(self.value)
+
+    def log(self, string):
+        if self.debug:
+            try:
+                print '[Viaplay]: %s' % string
```
Rechi: `print('[Viaplay]: %s' % string)` (if you can't use `xbmc.log(...)` here) for python 3 compatibility

plugin.video.viaplay/resources/lib/viaplay.py
```diff
@@ -0,0 +1,347 @@
...
+            except UnicodeEncodeError:
+                # we can't anticipate everything in unicode they might throw at
+                # us, but we can handle a simple BOM
+                bom = unicode(codecs.BOM_UTF8, 'utf8')
+                print '[Viaplay]: %s' % string.replace(bom, '')
```
Rechi: `print('[Viaplay]: %s' % string.replace(bom, ''))`

emilsvennesson: @Rechi Would you mind if I fix this in upcoming versions? I'm sure there is a lot more to take care of in regards to Python 3 compatibility.

romanvm: IIRC using `print` (no matter, Python 2 or 3 style) is not recommended for logging. All logging should be done with `xbmc.log()`

MartijnKaijser: Correct. Except for unit testing that would operate outside Kodi. 
If @emilsvennesson can update the PR or will do follow up to fix remaining items in a follow up PR i'd be ok with that to not stall the update too long.

plugin.video.viaplay/addon.py
```diff
@@ -3,601 +3,415 @@
...
+def channels_page(url):
+    channels_dict = helper.vp.get_channels(url)
+
+    for channel in channels_dict['channels']:
+        params = {
+            'action': 'list_products',
+            'url': channel['_links']['self']['href']
+        }
+        art = {
+            'thumb': channel['content']['images']['fallback']['template'].split('{')[0],
```
enen92: You could probably catch exceptions here. Some times there are no images returned by the REST API for some content and as a result your plugin will crash here when most likely it could still proceed to playback.

plugin.video.viaplay/resources/lib/viaplay.py
emilsvennesson: @romanvm @MartijnKaijser `viaplay.py` is supposed to be independent of Kodi, which is why I'm using `print` rather than importing any of the xbmc modules.

plugin.video.viaplay/addon.py
```diff
@@ -3,601 +3,415 @@
...
+
+    details = movie['content']
+
+    movie_info = {
+        'mediatype': 'movie',
+        'title': details['title'],
+        'plot': details.get('synopsis'),
+        'genre': ', '.join([x['title'] for x in movie['_links']['viaplay:genres']]),
+        'year': details['production'].get('year'),
+        'duration': int(details['duration'].get('milliseconds')) / 1000 if 'duration' in details else None,
```
enen92: In python3 dividing 2 integers will result in a float. You can use the `//` operator to ensure the result is always an integer. Not something you should worry right now, just pointing issues you'll have when moving the plugin to python3.

plugin.video.viaplay/addon.py
emilsvennesson: @enen92 I've thought about it, but their API always supplies generic images if there are none available.

plugin.video.viaplay/resources/lib/kodihelper.py
```diff
@@ -0,0 +1,228 @@
...
+        Attempts to authorize the device using the interval returned by the activation data."""
+        activation_data = self.vp.get_activation_data()
+        message = self.language(30039).format(activation_data['verificationUrl'], activation_data['userCode'])
+        dialog = xbmcgui.DialogProgress()
+        xbmc.sleep(200)  # small delay to prevent DialogProgress from hanging
+        dialog.create(self.language(30040), message)
+        secs = 0
+        expires = activation_data['expires']
+
+        while secs < expires:
```
enen92: You can use `xbmc.Monitor().abortRequested()` in your control loop statement to ensure it always breaks when kodi wants to exit

plugin.video.viaplay/resources/lib/viaplay.py
romanvm: Ok, but you should also use:
```python
from __future__ import print_function
```

plugin.video.viaplay/resources/lib/kodihelper.py
```diff
@@ -0,0 +1,228 @@
...
+                dialog.close()
+                return True
+            except self.vp.ViaplayError as error:
+                # raise all non-pending authorization errors
+                if not error.value == 'DeviceAuthorizationPendingError':
+                    raise
+            secs += activation_data['interval']
+            percent = int(100 * float(secs) / float(expires))
+            dialog.update(percent, message)
+            xbmc.sleep(activation_data['interval'] * 1000)
```
enen92: Same here, `xbmc.Monitor().waitForAbort(seconds)` is preferable

plugin.video.viaplay/resources/lib/viaplay.py
enen92: @emilsvennesson you should use the xbmc.log method instead of the print statement

plugin.video.viaplay/addon.py
enen92: Ah, don't worry about it then :)

plugin.video.viaplay/resources/lib/kodihelper.py
emilsvennesson: @enen92 I'm unsure on where I should implement the `waitforAbort` method. This is my loop with `abortRequested`:

```python
        while not xbmc.Monitor().abortRequested() and secs < expires:
            try:
                self.vp.authorize_device(activation_data)
                dialog.close()
                return True
            except self.vp.ViaplayError as error:
                # raise all non-pending authorization errors
                if not error.value == 'DeviceAuthorizationPendingError':
                    raise
            secs += activation_data['interval']
            percent = int(100 * float(secs) / float(expires))
            dialog.update(percent, message)
            xbmc.sleep(activation_data['interval'] * 1000)
            if dialog.iscanceled():
                dialog.close()
                return False

        dialog.close()
        return False
```

plugin.video.viaplay/addon.py
emilsvennesson: Definitely noted. Thanks! :-)

plugin.video.viaplay/resources/lib/viaplay.py
Rechi: from my side updating it isn't a hard requirement, just wanted to inform you that this won't work any more with python3

plugin.video.viaplay/resources/lib/kodihelper.py
enen92: Instead of xbmc.sleep basically. The sleep method should only be used for really small intervals

emilsvennesson: @enen92 Ah, got it!

emilsvennesson:
I have updated the PR now!

MartijnKaijser:
@emilsvennesson thx

PR #1602, opened 2018-01-14
[plugin.video.gronkh.de] 1.1.3

ToddShepard:
### Description
This updates the Gronkh.de plugin which currently is in the Gotham branch. The function which was lost due to API and TLS certificate changes is now restored. No features have been added or changed.

### Checklist:


MartijnKaijser:
Sorry but we actually only want to add new addons to krypton. Why did you pick jarvis?

ToddShepard:
Hey, I picked Jarvis because I believe that some people are using it, and it would be nice to have a working add-on there too instead of the broken version from Gotham. Alternatively I could also make the PR against Gotham, but I am not sure if I will be able to keep everything compatible in future releases.

MartijnKaijser:
@ToddShepard i can tell you even Jarvis/Krypton have issues with TLS on some platforms so you can't even guarantee it works there. Only Krypton we know works for sure.

ToddShepard:
That's why the TLS certificate check is disabled right now, it was planned to update the addon here with this PR to have a working legacy version and to add new features (and reenabling certificate check) than based on Krypton.

ToddShepard:
Okay, I have now opened a pull request for Gotham, see #1605

PR #1605, opened 2018-01-14
[plugin.video.gronkh.de] 1.1.3

ToddShepard:
### Description

This updates the Gronkh.de plugin which currently is in the Gotham branch. The function which was lost due to API and TLS certificate changes is now restored. No features have been added or changed.
### Checklist:


MartijnKaijser:
Thx
Could you fix the following?
PROBLEM: Fanart should have either 1280x720 or 1920x1080 but it has 3840x2160
It's a bit too big.

ToddShepard:
Fixed

PR #1779, opened 2018-04-09
Add travis buddy to Krypton

razzeee:
Travis buddy will comment on every PR on either failure or success of the ci and attach the log. That way authors don't need to leave github or search in travis.

I figured we can start with one branch take some time, maybe align it and then role it out to everywhere.

So this is mostly about a nice frendly message on success/failure.

@enen92 @Rechi @Lunatixz

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,5 +11,6 @@ script:
   - 'if [ "$TRAVIS_PULL_REQUEST" != "false" ]; then kodi-addon-checker $(git diff --diff-filter=d --name-only HEAD~ | grep / | cut -d / -f1 | sort | uniq); fi'
 
 notifications:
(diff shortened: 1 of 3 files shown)


Razzeee:
This above example is not what we will get with that markdown file. As travis buddy looks in the targeted branch for those files (no idea why, but that's how it is)

Rechi (approved):

enen92:
Thanks :)

Razzeee:
Let's see how it works before you thank me ;)
@mzfr was also involved big time

mzfr:
@Razzeee I think everything will be nice and good :)

Razzeee:
Hrm https://github.com/xbmc/repo-plugins/pull/1781#issuecomment-380597454

PR #1781, opened 2018-04-11
[plugin.video.francetv] 1.0.0

melmorabity:
### Description
This add-on provides access to the [France Télévisions](https://www.france.tv/) live and catchup TV platform.
### Checklist:

Additional information :


plugin.video.francetv/addon.py
```diff
@@ -0,0 +1,272 @@
+# Copyright (C) 2018 melmorabity
```
enen92: It is a general good practice to keep the entrypoint python file to a minimum since it won't get compiled to bytecode by the python interpreter. Usually you store the main file in the `resources` folder and just call it from the entrypoint:

```python
from resources.lib import addon

addon.run()
```

Do you want to take the oportunity to implement something like this and improve your addon performance or do you want to keep it for a future update?

Regards

plugin.video.francetv/addon.xml
```diff
@@ -0,0 +1,29 @@
...
+  <requires>
+    <import addon="xbmc.python" version="2.25.0"/>
+    <import addon="script.module.dateutil" version="2.5.3"/>
+    <import addon="script.module.requests" version="2.18.4"/>
+  </requires>
+  <extension point="xbmc.python.pluginsource" library="addon.py">
+    <provides>video</provides>
+  </extension>
+  <extension point="xbmc.addon.metadata">
+    <summary lang="en_GB">france•tv</summary>
```
enen92: Not sure this will work accross all skins (I guess it depends on the fontset) and not sure it works in our addon website. Who should I ping? @MartijnKaijser

plugin.video.francetv/addon.py
melmorabity: @enen92 Thaks for your comment, I will update the add-on with your suggestions.

plugin.video.francetv/addon.xml
melmorabity: @enen92 To avoid any potential problem, I will also update the add-on summary to avoid tofus with fonts without the "•" glyph.

enen92:
Thanks a lot for all the adjustments

PR #1794, opened 2018-04-16
[plugin.video.trailer.addict] broken

enen92:
Marking this addon as broken:

- Website changed
- None of the categories work
- Plugin is unmaintained (last update in 2015 and no activity in the forum)

Better marking it as broken (even if an update is submitted some day) than letting users install it in its current state.


Lunatixz:
@enen92 let me cook up a replacement for the Krypton branch.

PR #1809, opened 2018-04-24
[plugin.video.bcf] 1.0.0

FraserChapman:
### Description

This add-on enables playing of videos and movies from the [British Council Film](http://film.britishcouncil.org/british-council-film-collection) website.

The British Council Film Collection is an archive of 120 short documentary films made by the British Council during the 1940s designed to show the world how Britain lived, worked and played. 

This add-on is not created, maintained or in any way affiliated with the British Council.
It only provides an interface to access the free content on the British Council Film website from Kodi.

### Checklist:
- [x] Each add-on submission should be a single commit with using the following style: [plugin.video.bcf] v1.0.0

Additional information:

Follows almost the exact same format as my [last plugin](https://github.com/FraserChapman/plugin.video.bfi)  ...(it has a similar name too hence my screwing up that last pull request :( ).  Figured I'd reuse the code!


plugin.video.bcf/resources/lib/plugin.py
```diff
@@ -0,0 +1,260 @@
...
+from resources.lib import search as bcfs
+
+plugin = routing.Plugin()
+ADDON = xa.Addon()
+ADDON_ID = ADDON.getAddonInfo("id")  # plugin.video.bcf
+ADDON_NAME = ADDON.getAddonInfo("name")  # British Council Film Player
+MEDIA_URI = "special://home/addons/{}/resources/media/".format(ADDON_ID)
+
+BCF_THEMES = [
+    "Any",
```
enen92: would be nice to see these as localized strings in the future if someone wants to provide translations to your addon via transifex

plugin.video.bcf/resources/lib/plugin.py
```diff
@@ -0,0 +1,260 @@
...
+    "Second World War",
+    "Science and Technology",
+    "Scotland",
+    "Sport",
+    "Town and Country",
+    "Leisure",
+    "Landscapes and Scenery"]
+
+BCF_SERIES = [
+    "Any",
```
enen92: same

plugin.video.bcf/addon.xml
```diff
@@ -0,0 +1,33 @@
...
+        <provides>video</provides>
+    </extension>
+    <extension point="xbmc.addon.metadata">
+        <summary lang="en_GB">Search and play free movies from the British Council Film website</summary>
+        <description lang="en_GB">This add-on enables playing of videos and movies from the British Council Film website. It is not created, maintained or in any way affiliated with the British Council. The add-on only provides an interface to access the free content on the British Council Film website from Kodi.</description>
+        <language>en</language>
+        <platform>all</platform>
+        <license>MIT</license>
+        <forum></forum>
+        <website>www.fraserchapman.co.uk</website>
```
enen92: Just noticed this. This refers to the website of the content, can you change it (and the bfi player as well) in the next update? Thanks

FraserChapman: Ah - my mistake - will do!

enen92: @FraserChapman had a go with your plugins since I thought they are awesome to promote on social media. However got a few errors. On OSX and linux entering any section of your addon throws an exception. I tried to debug it and it looks like the sql connection on your cache module is always kept to None. This happens because when the addon is installed the profile folder does not exist. Hence, your plugin cannot create the sqlite file.
I managed to solve it by changing a plugin setting. This will create the profile folder and create a settings.xml file in the userdata/addon_data/yourplugin folder.
Running your plugin after the folder has been creating makes it work as expected.

You might want to check the existence of `xbmcaddon.Addon().getInfo("profile")` and create ithe folder if it does not exist before setting up the connection to the database in the Cache class constructor

Cheers

FraserChapman: @enen92 

Good catch! Didn't realise the "userdata/addon_data/[pluigin_id]" directory wasn't created on install - only when the settings have changed...on some platforms.

Oddly I think it does get created on Windows 10 (via the store app version) and did for me too on Mac OSX High Sierra - but  I get the same error on Debian :(

In any case it is a trivial fix - well ammed both and open pulls for 1.0.1 on both today/tomorrow if that is cool? Didn't want to really update so soon but a fix seems needed here,
Would also tweak the other bits you suggested (localised strings, etc) 

Thanks again for your diligence and time looking over these - if I get them perfected it would be great if you promoted them 😀 

Fraser

PR #1834, opened 2018-05-13
[plugin.video.dazn] 1.0.8

JinRonin:
upgrade to latest api version


plugin.video.dazn/resources/lib/common.py
```diff
@@ -55,10 +67,13 @@ def get_string(self, id_):
     def dialog_ok(self, id_):
         self.get_dialog().ok(self.addon_name, self.get_string(id_))
 
+    def notification(self, title, msg, duration, thumb):
+        xbmc.executebuiltin('XBMC.Notification({0}, {1}, {2}, {3})'.format(title, msg, duration, thumb))
```
enen92: please use the python api method instead of the builtin function:

`xbmcgui.Dialog().notification()`

JinRonin: True. I have removed it for now as its not being used yet anyway.

Thanks

Lunatixz:
LGTM

PR #1840, opened 2018-05-15
[plugin.video.jafc] 1.0.0

FraserChapman:
### Description

This add-on enables playing of videos and movies from the [Japanese Animated Film Classics](https://animation.filmarchives.jp/) website.

# About Japanese Animated Film Classics 

> This site was established to commemorate the 100th anniversary of Japanese animation, which is believed to have first been publicly shown in 1917. This site was developed with the National Institute of Informatics (NII) as part of the results of research on new methods of releasing collections, conducted as part of the "National Research Project for the Sustainability of Born-Digital Cinema (BDC Project)" under the 2016 "Cultural Arts Promotion Grant (Projects to Assist in the Promotion of Priority Areas in Art and History Museums)." It is hoped that this site will allow the collections of Japan's early animations preserved at NFC to reach a greater audience, be widely shared, to promote interest in and discovery of historical and cultural enjoyment, and to encourage contributions to creative work. 

# Disclaimer 

This add-on is not created, maintained or in any way affiliated with the National Film Archive of Japan.
It only provides an interface to access the free content on the Japanese Animated Film Classics website from Kodi.

### Checklist:
- [x] Each add-on submission should be a single commit with using the following style: [plugin.video.jafc] v1.0.0

Additional information :

Same format as my other plugins using the same basic framework. Took a little time to put this one together as I've been on my holls :) ...plus I didn't realise at first the site has an English version!


plugin.video.jafc/main.py
```diff
@@ -0,0 +1,5 @@
+# -*- coding: utf-8 -*-
+
+from resources.lib import plugin
+
```
Lunatixz: Can you insert `if __name__ == '__main__':` to avoid unwanted code execution.

plugin.video.jafc/resources/lib/search.py
```diff
@@ -0,0 +1,183 @@
...
+JAFC_SEARCH_URI = JAFC_URI + "en/works/"
+JAFC_INFO_URI = JAFC_URI + "/en/works/view/"
+JAFC_M3U8_TEMPLATE = "https://h10.cs.nii.ac.jp/stream/nfc/{}_en/hls/auto/media-2/stream.m3u8"
+
+
+SEARCH_SAVED = ku.get_setting_as_bool("search_saved")
+SEARCH_MAX_RESULTS = ku.get_setting_as_int("search_max_results")
+SEARCH_TIMEOUT = 60
+
+CACHE_URI = ku.translate_path("special://profile/addon_data/plugin.video.jafc/cache.sqlite")
```
Lunatixz: Please use API, 
```
import xbmcaddon
ADDON = xbmcaddon.Addon()
ADDON_DATA = ADDON .getAddonInfo('profile').decode('utf-8')
CACHE_URI = os.path.join(ADDON_DATA,'cache.sqlite')
```

plugin.video.jafc/main.py
Lunatixz: Please insert `if __name__ == '__main__':` to avoid unwanted code execution.

plugin.video.jafc/resources/lib/plugin.py
```diff
@@ -0,0 +1,300 @@
...
+from xbmcgui import ListItem
+
+from resources.lib import kodiutils as ku
+from resources.lib import search as jafc
+
+plugin = routing.Plugin()
+ADDON = xa.Addon()
+ADDON_ID = ADDON.getAddonInfo("id")  # plugin.video.jafc
+ADDON_NAME = ADDON.getAddonInfo("name")  # East Anglian Film Archive
+MEDIA_URI = "special://home/addons/{}/resources/media/".format(ADDON_ID)
```
Lunatixz: Please use API: https://codedocs.xyz/xbmc/xbmc/group__python__xbmcaddon.html

Lunatixz: ADDON_PATH  = ADDON_ID.getAddonInfo('path').decode('utf-8')
MEDIA_URI  = os.path.join(ADDON_PATH , 'resources', 'media','')

plugin.video.jafc/resources/lib/plugin.py
```diff
@@ -0,0 +1,300 @@
+# -*- coding: utf-8 -*-
+"""Main plugin file - Handles the various routes"""
+
+__author__ = "fraser"
+
+import routing
+import xbmc
+import xbmcaddon as xa
```
Lunatixz: It's good coding practice to leave these imports as they are named; This makes for easy code comprehension.

plugin.video.jafc/resources/lib/kodiutils.py
```diff
@@ -0,0 +1,119 @@
+# -*- coding: utf-8 -*-
+
+import json as json
```
Lunatixz: unnecessary `import json` is fine.

plugin.video.jafc/resources/lib/plugin.py
FraserChapman: Why on earth would I use os.path to join a special path URL - that makes no sense at all as the special path itself is path translated via xbmc.translatePath

plugin.video.jafc/resources/lib/search.py
FraserChapman: This seems arbitrary rather than objective - the other three plugins I have submitted that have been accepted use *exactly* the same code. Also - why on earth would I use os.path to join a SPECIAL url? That doesn't seem to make any sense.

plugin.video.jafc/main.py
FraserChapman: This seems an odd request - I have submitted and had accepted 3 plugins that don't have this, moreover on looking at other plugins in the repo almost none have this either. I can also see no mention of it in the plugin dev notes...

plugin.video.jafc/resources/lib/plugin.py
FraserChapman: This seems arbitrary rather than objective - the other three plugins I have submitted that have been accepted use *exactly* the same code.

plugin.video.jafc/resources/lib/kodiutils.py
FraserChapman: Will fix

FraserChapman:
Hmm - 90% of the requested changes seem a bit confused/arbitrary - will fix the unneeded import but would like some clarification / second opinion on the other bits please :)

plugin.video.jafc/resources/lib/plugin.py
Lunatixz: I wouldn't classify it as "arbitrary" submission rules do state we follow pep8 guidelines. I wouldn't hold back your submission because of this; I was merely letting you know so in the future you can keep it in mind. =)

Lunatixz:
90% seem arbitrary? I can assure you I used discretion when requesting changes. If you'd like a second opinion that's fine; someone else will review your code in the future.

enen92:
@FraserChapman do not understand PR review comments as mandatory changes. Must of the time they are only meant to enforce/suggest good practices. While I do not agree the `if __name__ == "__main__"` suggestion will make any noticeable difference nor the renaming of `xbmcaddon` to `xa` will make review harder, avoiding using the `special://` protocol would be good! You are inside the context of your addon when invoking `xbmcaddon.Addon()` and the object will hold references to the already translated paths you need to use, namely `path` (addon dir) and `profile` (addon_data). There is no reason to use the `special://` protocol and any path translation at all. 
Python is not sandboxed and kodi addons can pretty much run any code in the target machine. Addon developers, must of the time, just copy and paste code from other addons. The fact you use the `special://` path to access your own profile folder makes it look like any directory in userdata can be freely accessed from any addon which is wrong. When reviewing plugins, even if there isn't anything wrong with the code, we have to think ahead and avoid errors commited by others in the future.

That being said, there is absolutely nothing wrong with your code. It can be merged without those changes (well I'd like to look into the code too). However it is nice to warn about these kind of issues and seing them fixed/improved is always good.

FraserChapman:
@Lunatixz @enen92 - thanks for the feedback guys - sorry if I sounded cranky, was first thing pre coffee and I guess I was a bit confused after the same code had gone through before. 

I really appreciate the time and effort going through the code and totally didn't mean to sound dismissive or anything. After reading your feedback I will look to make the changes here and in the other plugins too - really I just want to make them as solid/usable as possible. 

Thanks - and sorry again I came across as a bit short :)

plugin.video.jafc/resources/lib/plugin.py
FraserChapman: Just looking at this again and it really doesn't make any sense. 
Your suggested change...

ADDON_PATH = ADDON_ID.getAddonInfo('path').decode('utf-8')

AttributeError: 'str' object has no attribute 'getAddonInfo'

plugin.video.jafc/resources/lib/plugin.py
FraserChapman: Hi,

which pep8 guideline says not to use aliases for imports?
Using pycharm and it doesn't flag it and I can't see any reference to this in the guide either...

https://www.python.org/dev/peps/pep-0008/#imports

Am I missing something?

FraserChapman:
Do I need to do something here to get the pull request accepted? After renewing the suggested changes they actually have errors in them - suggesting method calls on strings, etc. 
Don't mind making changes but the suggestions given are plain wrong...

Thanks :)

plugin.video.jafc/resources/lib/plugin.py
FraserChapman: Presume it would be 

ADDON_PATH = ADDON.getAddonInfo('path').decode('utf-8')

...

plugin.video.jafc/resources/lib/search.py
FraserChapman: When I try your suggested code I get.

File "/Users/fraser/Library/Application Support/Kodi/addons/plugin.video.jafc/resources/lib/cache.py", line 193, in _execute
                                                with self.connection:
                                            AttributeError: __exit__

When I do 

CACHE_URI = ku.translate_path(os.path.join(ADDON_DATA,'cache.sqlite'))

Then it works...really not sure how to implement your suggested change as it breaks the code :(

FraserChapman: FWIW 

os.path.join(ADDON_DATA,'cache.sqlite')  is equal to "special://profile/addon_data/plugin.video.jafc/cache.sqlite"

Lunatixz:
The suggestion are not "plain wrong"... If you implement them correctly they will work.

When I posted the code take it as pseudocode it was not written for you to copy and paste.

Please post your current broke code so I can see where you went wrong.

I also suggest you put a little more faith into the information being provided to you.

FraserChapman:
@Lunatixz - my code is actual code that works - and your pseudo code is wrong - i.e. produces errors, has typos, is incorrect. 

Not being funny but I'd suggest if you are going to review code and make suggestions then you take the time not to suggest broken pseudo code. 

I prefer rationality and evidence over faith :)

Lunatixz:
Paths are straight forward, how and why you are having problems adapting the API over your hard coded paths is unknown until you post the broken code. Please update your paths using the API and I'll review why they are not working for you.

plugin.video.jafc/resources/lib/plugin.py
enen92: Yes it should be a `xbmcaddon.Addon()` instance

plugin.video.jafc/resources/lib/plugin.py
enen92: Ignore this one

plugin.video.jafc/resources/lib/search.py
enen92: You don't need to execute translatepath on the paths provided by `xbmcaddon.Addon().getAddonInfo('profile')` or `xbmcaddon.Addon().getAddonInfo('path')`, they will return an already translated path. You only need to translate special:// paths. Your profile special path will be equal (after translation) to the one obtained from `getAddonInfo`. If they were not the same something was tremendously wrong :)

enen92:
@FraserChapman please see comments

FraserChapman:
@Lunatixz - if you have a working suggestion for my code please post it. Again my code - as is - 100% works, the only thing that doesn't and is broken is your "pesudo" code.  I don't really have time to decipher what you "might" mean, nor the inclination to guess. Also - if they are so easy I find it fairly odd that you can't seem to write two lines that actually work.

@enen92 - I've seen the comments, if they are simply suggestions then I'm happy with how it is at the moment. I'd be happy for the PR to be accepted or closed - which ever. 

What is bizarre is this is exactly the same as the 3 other plugins that were accepted and none of this was mentioned. I'm trying to contribute to an OS project - not play a guessing game or decipher ambiguous pseudo code.

enen92:
@FraserChapman the other 3 plugins like this one do not have any problem. Lunatixz has the opinion (like myself) that using special:// paths for obtaining paths that can be obtained directly from the python api is wrong. Others have request this change as mandatory in the past!
From the previous conversation I got the impression you'd update the PR to improve the mentioned comments. If you did not want to adjust or if you want to do that later just clearly state that. I think it was quite clear we were not requesting mandatory changes and the PR could be merged.

Regarding your last comment, we are not paid to review pull requests nor by github comment. We are also contributors. We invest a lot of our freetime with Kodi to guarantee PRs are reviewed and merged quickly. Maybe a little respect is deserved, no one is here to waste your time.

PR merged.

FraserChapman:
@enen92 - I hear you...but likewise I am not paid to try to write plugins either :) Respect is something earned it is never just deserved. Asking me to make changes with broken pseudo code is really disrespectful and is wasting my time as I fully tried to make the changes @Lunatixz  suggested but they are full of errors - some basic like calling methods on strings, etc. Also he/she goes on about the PEP8 style guide when actually the things suggested aren't in it (unless I am missing something). I guess it goes two ways, if you want people to contribute then having decent code reviews that don't suggest breaking changes, have basic errors, would be really helpful and way less disrespectful.

Lunatixz:
@FraserChapman If you were confused on how to implement my changes you could have asked. You were instantly dismissive of my suggestions. Which frankly IMO are basics!! All reviewers are different; some are stricter than others. You should understand there is a difference between "working" and "the right way". I mistakenly thought you would want to know the difference... in the end it's your sloppy code you'll have to maintain.

FraserChapman:
@Lunatixz - I was confused as to why you posted broken pseudo code, try to claim things are in the PEP guided that aren't, etc. The code you posted - for example - tries to call methods on strings - so basically wrong it is almost a joke. If you can't be bothered to get even the most basic things correct then why even review? It is less than helpful and is actually quite rude. 

Go back and actually look - every one of my responses is a question that you never addressed. For example - "Why on earth would I use os.path to join a SPECIAL url?" (...os.path.join is for cross platform path joining - but URLs always use a forward slash - so it is pointless). You still don't address them now - just try and act superior :)

Forgive me but I am usually dismissive when people do such things. FWIW I do post a lot on stackoverflow and love to help people - but if you answered questions like you review code then you would get nothing but downvotes :) have fun - in the end you have to be you.

plugin.video.jafc/resources/lib/search.py
FraserChapman: @enen92 - Sure that makes sense :) I think the confusion arose from the fact that the example given in the initial review was wrong.

Lunatixz:
Again, if my instructions confused you; you could have asked for help... I  felt I gave you enough information to guide you to a successful conclusion. Below are the two suggested changes, both function as expected.

plugin.py
```
import xbmcaddon, os
ADDON = xbmcaddon.Addon()
ADDON_LOC  = ADDON.getAddonInfo('path').decode('utf-8')
MEDIA_URI = os.path.join(ADDON_LOC,"resources","media")
```
search.py
```
import xbmcaddon, os
ADDON = xbmcaddon.Addon()
SETTINGS_LOC  = ADDON.getAddonInfo('profile')
CACHE_URI = ku.translate_path(os.path.join(SETTINGS_LOC,"cache.sqlite"))
```
Sometimes developers use the submission process to absorb new information; others don't. However there was never a reason for your level of disrespect and hostile tone.

enen92:
@Lunatixz @FraserChapman please move on. Harp on the same string won't get us anywhere. I've seen situations like this over and over again in the end we either lose a valid code contributor or a reviewer/team member. In each of those hypothetical situations only the project and our userbase will suffer the consequences.
I'm locking the conversation for now.

PR #1842, opened 2018-05-17
[plugin.video.fosdem] 0.0.1

jelly:
### Description
This is a video addon for Kodi which allows you to watch videos from FOSDEM 2018 event. The videos can be browsed through per day and per room. The plugin allows playback of either mp4 or webm videos which can be configured in the addon settings.

### Checklist:

- [x] Each add-on submission should be a single commit with using the following style: [plugin.video.fosdem] v1.0.0


jelly:
Fixed, thanks!

Rechi:
I think commit message was correct before, at least it is wrong now (correct: `[plugin.video.fosdem] v0.0.1`).

plugin.video.fosdem/addon.py
```diff
@@ -0,0 +1,132 @@
...
+
+    endOfDirectory(plugin.handle)
+
+
+@plugin.route('/event/<event_id>')
+def show_event(event_id):
+    event = root.find('.//event[@id="{}"]'.format(event_id))
+    videos = [link.attrib['href'] for link in event.findall('./links/link') if 'video.fosdem.org' in link.attrib['href']]
+    if not videos:
+        # XXX: How does this work?
```
enen92: you can set resolve url to False in `setResolvedUrl` arguments

plugin.video.fosdem/addon.xml
```diff
@@ -0,0 +1,30 @@
...
+        <provides>video</provides>
+    </extension>
+    <extension point="xbmc.addon.metadata">
+        <summary lang="en_GB">Videos of FOSDEM talks</summary>
+        <description lang="en_GB">This addon provides access to videos published on https://video.fosdem.org</description>
+        <language>en</language>
+        <platform>linux</platform>
+        <license>MIT</license>
+        <forum></forum>
+        <website>https://vdwaa.nl</website>
```
enen92: this is the website for the content

plugin.video.fosdem/addon.py
jelly: Sorry, I don't understand the comment, if I change the setResolvedUrl to setResolvedUrl(plugin.handle, False, ListItem(path=url)) the plugin no longer plays videos.

enen92: I mean in case there are no videos (you could not resolve the plugin path to a playable path) you can setResolvedUrl with the succeeded arg to False instead of just returning from your method. This is a minor thing, plugin can be merged without this.

PR #1888, opened 2018-06-11
[plugin.video.orftvthek] 0.8.4

s0faking:
### Description
reverted changes from 0.8.3 which caused kodi to crash when using the playall listitem
### Checklist:

Additional information :


Rechi:
@s0faking if it still crashes this needs a different fix, marking an item as folder which doesn't add items or marking it playable when it doesn't point to a media url is wrong.

s0faking:
@Rechi i agree but until we have a solution i want to prevent those kodi crashes which we have with the new fix. i suggest we work on that on the devel branch and release it when we have a stable version

Rechi:
If only v18 crashes, please let it crash.

s0faking:
17.6 is crashing. if it would have been v18 i certainly wouldnt have commited this temporary fix. i havent tested this with other version but since the current stable release is affected, i cant leave this as it is now until there is a working solution.  is it working with your kodi version? which one are you using?

Rechi:
newer version (0.8.5) submitted

PR #1890, opened 2018-06-12
[plugin.video.mlbtv] 2018.6.7

eracknaphobia:
### Description

        - Force Stream order in select popup (highlights, home, away) 
        - Fix bug that wouldn't allow exit of stream select popup
        - Add support for inputstream adaptive 
        - Fix for 60fps streams not available 
        - Move settings ID's to proper 30K range 
        - Add warning message when inputstream adaptive and always ask are both enabled 
### Checklist:

Additional information :


plugin.video.mlbtv/resources/lib/account.py
```diff
@@ -248,6 +248,14 @@ def get_stream(self, content_id):
         return stream_url, headers
 
     def get_stream_quality(self, stream_url):
+        #Check if inputstream adaptive is on, if so warn user and return master m3u8
+        if xbmc.getCondVisibility('System.HasAddon(inputstream.adaptive)'):
```
enen92: Unless the wiki has the wrong information regarding this boolean condition, I would suggest to use the json api instead and check for the *enable* state of the addon. This way you will trigger this message on every kodi install since `inputstream.adaptive` is bundled with kodi by default. JSON call:
```
{"jsonrpc":"2.0","method":"Addons.GetAddonDetails","params":{"addonid":"inputstream.adaptive","properties":["enabled"]},"id":4}
```

plugin.video.mlbtv/resources/lib/mlb.py
```diff
@@ -236,6 +238,7 @@ def stream_select(game_pk):
     else:
         xbmcplugin.setResolvedUrl(addon_handle, False, xbmcgui.ListItem())
         xbmc.executebuiltin('Dialog.Close(all,true)')
```
enen92: why do you need to call this?

eracknaphobia: It was my attempt to silence the "No playable stream found" on a setResolvedURL if the user hit cancel on the select dialog. I don't think it does what I wanted, so I can remove it.

plugin.video.mlbtv/resources/lib/account.py
eracknaphobia: The if statement does return false when inputstream is installed but not enabled. Same if is currently used in PS Vue addon. Not saying it's proper but it does work as expected.

enen92: Nice to know, we need to update the wiki then :)

plugin.video.mlbtv/resources/lib/mlb.py
enen92: Yes please remove for merging

PR #1905, opened 2018-06-18
[plugin.video.massengeschmack] 2.4.1

phoerious:
### Checklist:

Additional information :


plugin.video.massengeschmack/resources/lib/listing.py
```diff
@@ -56,19 +56,19 @@ def show(self):
         if 'true' == ADDON.getSetting('advanced.viewmodeFix'):
             self.setViewMode(510)
         else:
-            self.setViewMode(55)
+            self.setViewMode(self.__source.getViewMode())
```
enen92: Sorry this is not allowed according to our rules (https://kodi.wiki/view/Add-on_rules). I am not sure how it got unnoticed before but plugins should not force viewtypes on the skin. They should set the mediatype on the listitem and contenttype via xbmcplugin. Views are provided by the skin to each specific media type, the user is then able to change them to their taste.

phoerious: Would it be okay if I added a setting so the user can turn off this behaviour? Not having correctly set view modes is really annoying, but if the user wants, they can still turn it off (BTW the add-on *does* set the content type, but it's not enough to get nice-looking and consistent listings).

MartijnKaijser: it's pointless cause each skin has different IDs.

phoerious: I know. But at least for Estuary it looks decent. For every other skin you can turn it off. For Aeon Nox there is already a setting.

phoerious: Yes? No? Using this add-on without consistent view modes is really ugly and confusing.

enen92: @phoerious sorry for the delay. Can you make it as a setting (that can be disabled) and only set the viewmode if the user is using estuary? You can detect the skin via `xbmc.getSkinDir()`. Thanks

phoerious: I added a setting. This got closed somehow, so here's the new PR: #1949

PR #1950, opened 2018-07-16
[plugin.video.embycon] 1.5.30

faush01:
### Description

July release:
- Fix a few bugs
- Remove some redundant code
- Refactor some code for performance

### Checklist:

Additional information :


plugin.video.embycon/resources/lib/context_monitor.py
```diff
@@ -0,0 +1,56 @@
...
+    stop_thread = False
+
+    def run(self):
+
+        item_id = None
+        log.debug("ContextMonitor Thread Started")
+        context_up = False
+        is_embycon_item = False
+
+        while not xbmc.abortRequested and not self.stop_thread:
```
enen92: `xbmc.Monitor().abortRequested()` please

plugin.video.embycon/resources/lib/functions.py
```diff
@@ -936,27 +988,61 @@ def searchResults(params):
...
+            # set content type
+            xbmcplugin.setContent(handle, content_type)
+
+            dir_items, detected_type = processDirectory(details_result, progress, params)
+            if dir_items is not None:
+                xbmcplugin.addDirectoryItems(handle, dir_items)
+                xbmcplugin.endOfDirectory(handle, cacheToDisc=False)
+
+        elif not query_string:
+            xbmcgui.Dialog().ok("No Matches", "No items match your search.")
```
enen92: minor: can you move this to localized strings?

plugin.video.embycon/resources/lib/menu_functions.py
```diff
@@ -853,13 +853,13 @@ def showWidgets():
...
 
-def showSearch():
+def show_search():
     addMenuDirectoryItem(i18n('movies'), 'plugin://plugin.video.embycon/?mode=NEW_SEARCH&item_type=Movie')
     addMenuDirectoryItem(i18n('tvshows'), 'plugin://plugin.video.embycon/?mode=NEW_SEARCH&item_type=Series')
     addMenuDirectoryItem(i18n('episodes'), 'plugin://plugin.video.embycon/?mode=NEW_SEARCH&item_type=Episode')
     addMenuDirectoryItem("Song", 'plugin://plugin.video.embycon/?mode=NEW_SEARCH&item_type=Audio')
     addMenuDirectoryItem("Album", 'plugin://plugin.video.embycon/?mode=NEW_SEARCH&item_type=MusicAlbum')
-    #addMenuDirectoryItem("Artist", 'plugin://plugin.video.embycon/?mode=NEW_SEARCH&item_type=MusicArtists')
+    addMenuDirectoryItem("Person", 'plugin://plugin.video.embycon/?mode=NEW_SEARCH&item_type=Person')
```
enen92: Same with these items if possible

plugin.video.embycon/resources/lib/context_monitor.py
faush01: This was picked up in the initial submission as well, in the other service loop I have, I did try to switch over at that time but found a bug in Kodi
Notes from my last attempt to use the monitor.

# monitor.abortRequested() is causes issues, it currently triggers for all addon cancelations which causes
# the service to exit when a user cancels an addon load action. This is a bug in Kodi.
# I am switching back to xbmc.abortRequested approach until kodi is fixed or I find a work around

Here is the forum thread on it
https://forum.kodi.tv/showthread.php?tid=319161&highlight=abortRequested

I am not sure this was pack ported to Kodi 17 yet however to maintain backward compatibility with all version of 17.x of Kodi I think I should stick with my current implementation for now and switch to the new one in Kodi 18.

Is that ok?

plugin.video.embycon/resources/lib/functions.py
faush01: will do

plugin.video.embycon/resources/lib/menu_functions.py
faush01: will do

plugin.video.embycon/resources/lib/context_monitor.py
faush01: @enen92 
Just wondering if this was ok.
Having to support multiple version of Kodi 17 with the same code base is my priority now.
I am not even sure how I would support addons for different versions of Kodi 17.x

enen92: Oh ok I am using v18 for a while and I was not aware of this issue in Krypton. Please make sure you adapt if you add this addon to the Leia repo at some point. Thanks

PR #1979, opened 2018-08-08
[plugin.video.mediathekview] 0.5.0

tuxpoldo:
### Description
- Fixed error that prevented some shows from being downloaded
- Support for old Maria DB / MySQL versions
- Database downloads will be deleted right after processing
- "Play with subtitles" as context menu operation or automatically when configured
- Last searches
- Faster list of live streams
- Possibility to download a show as movie or as series episode (different download location, naming schema and NFO file)
- Fixed a bug that prevented the creation of the SQLite DB on some systems
- More efficient database structure
- New "mvupdate" options --force and --intervall
- More customisation of NFO-Files
- Automatic detection of season and episode
- Improved download settings
- "Update on Start" did not initiate an update on system start if no database was found
- Detailed database information shown in the log file
### Checklist:

Additional information :

### DIFF
--- a/plugin.video.mediathekview/resources/lib/downloader.py
+++ b/plugin.video.mediathekview/resources/lib/downloader.py
@@ -1,0 +1,361 @@
+# -*- coding: utf-8 -*-
+"""
+Download handler module
+
+Copyright (c) 2018, Leo Moll
+Licensed under MIT License
+"""
+
+# pylint: disable=import-error
+# pylint: disable=mixed-indentation, bad-whitespace, bad-continuation, missing-docstring
+
+# -- Imports ------------------------------------------------
+from __future__ import unicode_literals
+
+import os
+import re
+
+from contextlib import closing
+
+import xbmc
+import xbmcvfs
+
+import resources.lib.mvutils as mvutils
+
+from resources.lib.kodi.KodiUI import KodiBGDialog
+from resources.lib.filmui import FilmUI
+from resources.lib.ttml2srt import ttml2srt
+
+# -- Classes ------------------------------------------------
+class Downloader( object ):
+	def __init__( self, plugin ):
+		self.plugin		= plugin
+		self.database	= plugin.database
+		self.settings	= plugin.settings
+		self.notifier	= plugin.notifier
+
+	def play_movie_with_subs( self, filmid, only_set_resolved_url ):
+		film = self.database.RetrieveFilmInfo( filmid )
+		if film is None:
+			self.notifier.ShowError( 30990, self.plugin.language( 30991 ) )
+			return
+		ttmname = os.path.join( self.plugin.datapath, 'subtitle.ttml' )
+		srtname = os.path.join( self.plugin.datapath, 'subtitle.srt' )
+		subs = []
+		if self.download_subtitle( film, ttmname, srtname, 'subtitle' ):
+			subs.append( srtname )
+		( videourl, listitem ) = FilmUI( self.plugin ).get_list_item( None, film )
+		if listitem:
+			if subs:
+				listitem.setSubtitles( subs )
+			if only_set_resolved_url:
+				self.plugin.setResolvedUrl( True, listitem )
+			else:
+				xbmc.Player().play( videourl, listitem )
+
+	def download_subtitle( self, film, ttmname, srtname, filename ):
+		ret = False
+		if film.url_sub:
+			bgd = KodiBGDialog()
+			bgd.Create( 30978, filename + u'.ttml' )
+			# pylint: disable=broad-except
+			try:
+				bgd.Update( 0 )
+				mvutils.url_retrieve_vfs( film.url_sub, ttmname, bgd.UrlRetrieveHook )
+				try:
+					ttml2srt( xbmcvfs.File( ttmname, 'r' ), xbmcvfs.File( srtname, 'w' ) )
+					ret = True
+				except Exception as err:
+					self.plugin.info( 'Failed to convert to srt: {}', err )
+				bgd.Close()
+			except Exception as err:
+				bgd.Close()
+				self.plugin.error( 'Failure downloading {}: {}', film.url_sub, err )
+		return ret
+
+	def download_movie( self, filmid, quality ):
+		if not self.test_download_path( self.settings.downloadpathmv ):
+			return
+		film = self.database.RetrieveFilmInfo( filmid )
+		if film is None:
+			return
+		( filmurl, suffix, extension, ) = self.get_film_url_and_extension( film, quality )
+		# try to create a good name for the downloaded file
+		namestem = mvutils.cleanup_filename( film.title )[:80]
+		if not namestem:
+			# try to take the show name instead...
+			namestem = mvutils.cleanup_filename( film.show )[:64]
+			if not namestem:
+				namestem = u'Film'
+			namestem = namestem + '-{}'.format( film.id )
+		elif self.settings.movienamewithshow:
+			showname = mvutils.cleanup_filename( film.show )[:64]
+			if showname:
+				namestem = showname + ' - ' + namestem
+		# review name
+		if self.settings.reviewname:
+			( namestem, confirmed ) = self.notifier.GetEnteredText( namestem, 30986 )
+			namestem = mvutils.cleanup_filename( namestem.decode('utf-8') )
+			if len( namestem ) < 1 or confirmed is False:
+				return
+		# build year postfix
+		year = self.matches( '([12][0-9][0-9][0-9])', str( film.aired ) )
+		if year is not None:
+			postfix = ' (%s)' % year
+		else:
+			postfix = ''
+		# determine destination path and film filename
+		if self.settings.moviefolders:
+			pathname = self.settings.downloadpathmv + namestem + postfix + '/'
+			filename = namestem + suffix
+		else:
+			pathname = self.settings.downloadpathmv
+			filename = namestem + postfix + suffix
+		# check for duplicate
+		while xbmcvfs.exists( pathname + filename + extension ):
+			( filename, confirmed ) = self.notifier.GetEnteredText( filename, 30987 )
+			filename = mvutils.cleanup_filename( filename.decode('utf-8') )
+			if len( filename ) < 1 or confirmed is False:
+				return
+
+		# download the stuff
+		if self.download_files( film, filmurl, pathname, filename, extension ):
+			self.make_movie_nfo_file( film, filmurl, pathname, filename )
+
+	def download_episode( self, filmid, quality ):
+		if not self.test_download_path( self.settings.downloadpathep ):
+			return
+		film = self.database.RetrieveFilmInfo( filmid )
+		if film is None:
+			return
+
+		( filmurl, suffix, extension, ) = self.get_film_url_and_extension( film, quality )
+
+		# detect season and episode
+		( season, episode, fninfo, ) = self.season_and_episode_detect( film )
+
+		# determine names
+		showname = mvutils.cleanup_filename( film.show )[:64]
+		namestem = mvutils.cleanup_filename( film.title )[:80]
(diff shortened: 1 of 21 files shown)


plugin.video.mediathekview/resources/lib/downloader.py
```diff
@@ -0,0 +1,361 @@
...
+		self.database	= plugin.database
+		self.settings	= plugin.settings
+		self.notifier	= plugin.notifier
+
+	def play_movie_with_subs( self, filmid, external ):
+		film = self.database.RetrieveFilmInfo( filmid )
+		if film is None:
+			self.notifier.ShowError( 30990, self.plugin.language( 30991 ) )
+			return
+		ttmname = self.plugin.datapath + '/subtitle.ttml'
```
enen92: it would be nice to use the os module to join paths

plugin.video.mediathekview/resources/lib/downloader.py
```diff
@@ -0,0 +1,361 @@
...
+		srtname = self.plugin.datapath + '/subtitle.srt'
+		subs = []
+		if self.download_subtitle( film, ttmname, srtname, 'subtitle' ):
+			subs.append( srtname )
+		( videourl, listitem ) = FilmUI( self.plugin ).get_list_item( None, film )
+		if listitem:
+			if subs:
+				listitem.setSubtitles( subs )
+			if external:
+				xbmc.Player().play( videourl, listitem )
```
enen92: why you need to call `xbmc.Player()`? what is external exactly? An external player?
Note that calling xbmc.Player in plugins is a bad practice, an entry in the db will be created each time a user plays a file. Plugin:// paths, on the other hand, are unique.

plugin.video.mediathekview/resources/lib/filmui.py
```diff
@@ -27,96 +39,146 @@ def Begin( self, showshows, showchannels ):
...
+			info_labels['duration'] = film.seconds
+
+		if film.aired is not None:
+			airedstring = '%s' % film.aired
+			if airedstring[:4] != '1970':
+				info_labels['date']		= airedstring[8:10] + '-' + airedstring[5:7] + '-' + airedstring[:4]
+				info_labels['aired']		= airedstring
+				info_labels['dateadded']	= airedstring
+
+		icon = 'special://home/addons/' + self.plugin.addon_id + '/resources/icons/' + film.channel.lower() + '-m.png'
```
enen92: `xbmcaddon.Addon().getAddonInfo("icon")`

tuxpoldo: This is not correct. According to [the doc](https://codedocs.xyz/AlwinEsch/kodi/group__python__xbmcaddon.html#ga65a28456461e069aeeddd13a8a4222fb) `xbmcaddon.Addon().getAddonInfo("icon")` would return the path to the addon icon as specified in the `addon.xml` file (`<icon>resources/icon.png</icon>`) but in case of channel ARD the icon would be `ADDONPATH/resources/icons/ard-m.png`.

plugin.video.mediathekview/resources/lib/downloader.py
tuxpoldo: The routine `def play_movie_with_subs( self, filmid, external ):` has a double purpose. But first let me explain the circumstances:

The german public broadcasters have a URL with a (playable) media content (usually MP4). In addition many shows have also an additional URL for subtitles. Unfortunately these subtitles are not playable, since they are provided in an _"unusual"_ format that - as far as I know - is not supported by Kodi (TTML and in rare cases VTT).

If you want to play such a stream with subtitles, you have to download the TTML subtitle **BEFORE** playing the stream, convert it, create a new list item and **THAN** play the stream.

Regarding playing shows with subtitles there are two operating modes in the plugin, in dependency of a configuration setting.

If "Autoplay Subtitles" is on, the following happens: when the directory list is filled, each movie with subtitles will not be added with his real media url, but with the plugin URL itself and as parameters `{'mode': "playwithsrt", 'id': <MOVIEID>, 'external': False}`. This is a kind of deferred media URL retrieval mechanism ([See this forum post](https://forum.kodi.tv/showthread.php?tid=173986)). When the user selects the entry in the directory for playing, Kodi notices that the media URL is not a playable video URL and than invokes the specified plugin URL expecting the real URL to be resolved. Here the Plugin passes the supplied parameters to `play_movie_with_subs(<MOVIEID>, False)`. The method now downloads the incompatible subtitle file locally, converts it to a supported format fills in the information into a `listitem` and returns the information to Kodi by invoking `xbmcplugin.setResolvedUrl( True, listitem )` ([See Documentation](https://codedocs.xyz/AlwinEsch/kodi/group__python__xbmcplugin.html#gaa03ad528583db7d5277eae3977acde0f)). Kodi starts to play the media with the subtitles.

If "Autoplay Subtitles" is off, the following happens: when the directory list is filled, each movie will be added with his real media URL but with an additional context menu entry (_"Play with Subtitles"_). When the user specifically selects the context menu item _"Play with Subtitles"_, the Plugin itself will be invoked with the parameters `{'mode': "playwithsrt", 'id': <MOVIEID>, 'external': True}`. Here the Plugin passes the supplied parameters to `play_movie_with_subs(<MOVIEID>, True)`. The method now downloads the incompatible subtitle file locally, converts it to a supported format fills in the information into a `listitem` and than invokes `xbmc.Player().play( videourl, listitem )` because this is the only way known to me for playing an item from a command posted into the plugin.

If you have any better idea, please let me know. I invested quite a lot of time in finding this solution but I will be very glad to implement anything better, if possible.

plugin.video.mediathekview/resources/lib/downloader.py
tuxpoldo: You are right. Will fix it.

tuxpoldo:
@enen92 : I updated the pull request after following some of your suggestions

tuxpoldo:
The routine `def play_movie_with_subs( self, filmid, external ):` has a double purpose. But first let me explain the circumstances:

The german public broadcasters have a URL with a (playable) media content (usually MP4). In addition many shows have also an additional URL for subtitles. Unfortunately these subtitles are not playable, since they are provided in an _"unusual"_ format that - as far as I know - is not supported by Kodi (TTML and in rare cases VTT).

If you want to play such a stream with subtitles, you have to download the TTML subtitle **BEFORE** playing the stream, convert it, create a new list item and **THAN** play the stream.

Regarding playing shows with subtitles there are two operating modes in the plugin, in dependency of a configuration setting.

If "Autoplay Subtitles" is on, the following happens: when the directory list is filled, each movie with subtitles will not be added with his real media url, but with the plugin URL itself and as parameters `{'mode': "playwithsrt", 'id': <MOVIEID>, 'external': False}`. This is a kind of deferred media URL retrieval mechanism ([See this forum post](https://forum.kodi.tv/showthread.php?tid=173986)). When the user selects the entry in the directory for playing, Kodi notices that the media URL is not a playable video URL and than invokes the specified plugin URL expecting the real URL to be resolved. Here the Plugin passes the supplied parameters to `play_movie_with_subs(<MOVIEID>, False)`. The method now downloads the incompatible subtitle file locally, converts it to a supported format fills in the information into a `listitem` and returns the information to Kodi by invoking `xbmcplugin.setResolvedUrl( True, listitem )` ([See Documentation](https://codedocs.xyz/AlwinEsch/kodi/group__python__xbmcplugin.html#gaa03ad528583db7d5277eae3977acde0f)). Kodi starts to play the media with the subtitles.

If "Autoplay Subtitles" is off, the following happens: when the directory list is filled, each movie will be added with his real media URL but with an additional context menu entry (_"Play with Subtitles"_). When the user specifically selects the context menu item _"Play with Subtitles"_, the Plugin itself will be invoked with the parameters `{'mode': "playwithsrt", 'id': <MOVIEID>, 'external': True}`. Here the Plugin passes the supplied parameters to `play_movie_with_subs(<MOVIEID>, True)`. The method now downloads the incompatible subtitle file locally, converts it to a supported format fills in the information into a `listitem` and than invokes `xbmc.Player().play( videourl, listitem )` because this is the only way known to me for playing an item from a command posted into the plugin.

If you have any better idea, please let me know. I invested quite a lot of time in finding this solution but I will be very glad to implement anything better, if possible.

Maybe the name of the parameter `external` is not very self-explanatory. In the last change of the pull request, it was renamed to `only_set_resolved_url`

plugin.video.mediathekview/resources/lib/filmui.py
enen92: Disregard (though it was the addon icon), please use `os.path.join(xbmcaddon.Addon().getAddonInfo("path"), resources, icons, film.channel.lower() + '-m.png')` instead. Main reasoning is to use the python api instead of using the `special://` paths

tuxpoldo: OK - will fix.

tuxpoldo:
@enen92 - Pull request just updated again

Rechi:
> Each add-on submission should be a single commit with using the following style: [plugin.video.foo] v1.0.0

Please update your PR.

plugin.video.mediathekview/resources/lib/downloader.py
enen92: Yeah Kodi lacks support for ttml based subtitles (although inputstream.adaptive already has this functionality for mpd and hls manifests). Definitely something to look for in the future.
The explanation looks good to me and I can see your efforts in doing the logic properly. I am wondering, in the second case where you call `xbmc.Player().play()`, why not to simply use a builtin-function with the `plugin://` path instead of this class? 
Example:
`xbmc.executebuiltin("PlayMedia(plugin://yourplugin/arg/)")`

Another option could be to use the JSON-RPC (xbmc.executeJSONRPC with a `Player.Open` action). Would this work for you?

Rechi:
I know that you initially only had one commit, but if you do changes you should amend them instead of creating separate commits. Btw adding @ mentions is quite annoying for the person, as this will generate a notification every time a user pushes the commit on GitHub.

To fix your pull request execute the following commands (origin is your repo, upstream is the xbmc repo)

```
git fetch upstream
git checkout krypton
git reset --hard upstream/krypton
git checkout origin/krypton -- plugin.video.mediathekview/
git commit -m '[plugin.video.mediathekview] 0.5.0'
git push -f origin krypton
```

plugin.video.mediathekview/resources/lib/downloader.py
tuxpoldo: Thank you for the suggestions! The `xbmc.executebuiltin("PlayMedia(plugin://yourplugin/arg/)")` sounds very promising. I will give it a try and if it works, I will implement it in one of the next versions.

tuxpoldo:
Thank you for the explanation how to squash the commits in the originating repository. This was really helpful and I will follow this method in the next PR. In the meantime I tried the solution described in [this article](https://stackoverflow.com/questions/5189560/squash-my-last-x-commits-together-using-git), but it seemed not to be working....

plugin.video.mediathekview/resources/lib/downloader.py
tuxpoldo: Tracked Issue in https://github.com/mediathekview/plugin.video.mediathekview/issues/88

tuxpoldo:
The problem is, that 'git fetch upstream' does not work. The result is

````
fatal: 'upstream' does not appear to be a git repository
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
````

I suppose that `upstream` is *THIS* repository. I'll make a tentative by adding:

````
git remote add upstream git@github.com:xbmc/repo-plugins.git
````

Rechi:
> origin is your repo, upstream is the xbmc repo

Yes, you have the right command (git remote add)

tuxpoldo:
Done. It seems to work. The PR is now one squashed commit.

PR #2005, opened 2018-08-28
[plugin.video.watchmojo] 3.0.2

willforde:
### Description
Fixed some scraping issues that added some useless entries.
Move main code into a module, makes for a small entry point.

### Checklist:


plugin.video.watchmojo/addon.py
```diff
@@ -17,156 +17,12 @@
...
-    :param Resolver plugin: Tools related to Resolver callbacks.
-    :param unicode url: The url to a video.
-    :return: A playable video url.
-    """
-    url = url_constructor(url)
-    return plugin.extract_source(url)
+import sys
+import os
 
+lib = os.path.join(os.path.dirname(__file__), "resources", "lib")
```
enen92: please properly turn this into a python module by creating `__init__` files along the paths you need to import. `sys.path.append` is not a clean way to import your dependencies.

plugin.video.watchmojo/resources/__init__.py
```diff
@@ -0,0 +1 @@
+pass
```
enen92: you don't need `pass`, empty files will do it

plugin.video.watchmojo/resources/lib/__init__.py
```diff
@@ -0,0 +1 @@
+pass
```
enen92: same

willforde:
O, good to know. I might as well make the change. Thanks.

PR #2043, opened 2018-09-11
[plugin.video.dazn] 1.2.1-broken

JinRonin:
Marked as broken as its not compatible with krypton anymore.
New version for leia will be pushed later on.


Rechi:
@JinRonin version has to be bumped (higher then 1.2.0, but lower then 1.3.0 to not break the leia version) otherwise it doesn't get updated.

Rechi:
@JinRonin now the version in the commit message and addon.xml don't match anymore.

PR #2058, opened 2018-09-22
[plugin.video.infowars] 2.0.8

Spinalcracker:
Version 2.0.8
- Removed Censored Content (Spinalcracker)
- Cleaned up Menu System
- Added New Shows
- Added Millie Weaver Youtube Channel

### DIFF
--- a/plugin.video.infowars/default.py
+++ b/plugin.video.infowars/default.py
@@ -12,7 +12,7 @@
 ### ############################################################################################################
 ##### Imports #####
 import xbmcplugin, xbmcgui, xbmcaddon, xbmcvfs, xbmc
-import urllib,urllib2,re,os,sys,htmllib,string,StringIO,logging,random,array,time,datetime
+import urllib,urllib2, urllib3, re,os,sys,htmllib,string,StringIO,logging,random,array,time,datetime, ssl, socket
 import copy
 import HTMLParser, htmlentitydefs
 try: 		from sqlite3 										import dbapi2 as sqlite; print "Loading sqlite3 as DB engine"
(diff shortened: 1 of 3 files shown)


plugin.video.infowars/default.py
```diff
@@ -354,65 +364,71 @@ def Menu_MainMenu(): #The Main Menu
...
+        content = response.data
         videos= find_multiple_matches(content,"<ul class=\"show__footer\"(.*?)</ul>")
         sources = []
         for temp, entry in enumerate(videos): 
             if temp == 0:
                 video_id = find_single_match(entry,"data-youtube-src=\"https://www.youtube.com/embed/(.*?)\"")
                 url = "plugin://plugin.video.youtube/play/?video_id=%s" % video_id
                 #IW_addon.log('SPINAL VIDEO_ID IS %s' % video_id)
-                PlayURL(url)
+                xbmc.Player().play(url)
```
enen92: please don't use xbmc.Player().play() from plugins. This will create a new entry in the database each time a file is played. In plugins you should `setProperty("isPlayable", "true")` in the listitem and route the item to `xbmcplugin.setResolvedUrl()`. This way you request the playback of a `plugin://` path which is unique.

plugin.video.infowars/default.py
```diff
@@ -354,65 +364,71 @@ def Menu_MainMenu(): #The Main Menu
...
+    if response:
+        content = response.data
         videos= find_multiple_matches(content,"<ul class=\"show__footer\"(.*?)</ul>")
         sources = []
         for temp, entry in enumerate(videos): 
             if temp == 1:
                 video_id = find_single_match(entry,"data-youtube-src=\"https://www.youtube.com/embed/(.*?)\"")
                 url = "plugin://plugin.video.youtube/play/?video_id=%s" % video_id
-                PlayURL(url)
+                play=xbmc.Player() ### xbmc.PLAYER_CORE_AUTO | xbmc.PLAYER_CORE_DVDPLAYER | xbmc.PLAYER_CORE_MPLAYER | xbmc.PLAYER_CORE_PAPLAYER
```
enen92: same

plugin.video.infowars/default.py
```diff
@@ -354,65 +364,71 @@ def Menu_MainMenu(): #The Main Menu
...
+    if response:
+        content = response.data
         videos= find_multiple_matches(content,"<ul class=\"show__footer\"(.*?)</ul>")
         sources = []
         for temp, entry in enumerate(videos): 
             if temp == 2:
                 video_id = find_single_match(entry,"data-youtube-src=\"https://www.youtube.com/embed/(.*?)\"")
                 url = "plugin://plugin.video.youtube/play/?video_id=%s" % video_id
-                PlayURL(url)
+                play=xbmc.Player() ### xbmc.PLAYER_CORE_AUTO | xbmc.PLAYER_CORE_DVDPLAYER | xbmc.PLAYER_CORE_MPLAYER | xbmc.PLAYER_CORE_PAPLAYER
```
enen92: same

plugin.video.infowars/default.py
Spinalcracker: same

Lines 373 - 431 are commented out, old code, left in for reference for the other addon maintainer, but not actually executed.  I can delete that whole block and repull, but thought this would be ok.

plugin.video.infowars/default.py
Spinalcracker: same

Lines 373 - 431 are commented out, old code, left in for reference for the other addon maintainer, but not actually executed.  I can delete that whole block and repull, but thought this would be ok.

plugin.video.infowars/default.py
Rechi: Previous code can be looked up in the version code system you use for developing the add-on.

Spinalcracker: Ya for sure, I was being lazy.  Removing now.

Spinalcracker:
Closing, making changes and will re-pull

PR #2082, opened 2018-10-03
[plugin.video.tweakers] 1.1.10

skipmodea1:
### Description
v1.1.10 (2018-10-01)
- removing non-ascii characters in title in parameters to prevent UnicodeDecodeError: 'ascii' codec can't decode ...
The error occured when doing urllib.parse.parse_qs of the parameters
- using youtube now (website change)
- removed video quality setting as the youtube addon is used
### Checklist:

### DIFF
--- a/plugin.video.tweakers/addon.xml
+++ b/plugin.video.tweakers/addon.xml
@@ -2,20 +2,20 @@
 <addon 
 	id="plugin.video.tweakers" 
 	name="Tweakers" 
-	version="1.1.9"
+	version="1.1.10"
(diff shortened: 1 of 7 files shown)


skipmodea1:
Why am i getting 'ERROR: Required addon plugin.video.youtube not available in current repository.'? And how do i fix that?

Rechi:
Because it's not available in gotham. Either remove it as dependency or submit to a higher version.

skipmodea1:
Right, thanks!

PR #2116, opened 2018-10-17
[plugin.video.telebasel] 0.9.0

goggle:
### Description

Telebasel is a local Swiss TV station (from Basel). This Kodi plugin (plugin.video.telebasel) provides a way to access its media library (https://telebasel.ch/mediathek) through Kodi.

### Checklist:

Additional information :


plugin.video.telebasel/resources/lib/telebasel.py
```diff
@@ -0,0 +1,537 @@
...
+            # The element 'Telebasel Archiv' refers to the old
+            # website. We need to skip this:
+            if 'Archiv' in title:
+                continue
+
+            shows.append({'title': title, 'page': page, 'image': image_url})
+
+        if extract_channel_id:
+            feed_urls = [x['page'] + '&podcast' for x in shows]
+            pool = ThreadPool(len(feed_urls))
```
enen92: last time I checked multiprocessing is prone to errors in some platforms (don't remember exactly which), maybe its solved by now. In how many different platforms have you tested your addon?
Normally devs try to implement a fallback in case multiprocessing fails, see: https://github.com/xbmc/repo-plugins/blob/krypton/plugin.video.locast/resources/lib/locast.py

Maybe @Lunatixz can explain a bit better where it is prone to fail

Lunatixz: I believe this issue is limited to Kodi variants, ie Openelec, LibreELEC which ship lightweight python packages. Please include a fallback to Threadpool for incompatible systems.

goggle:
@enen92 Thanks for the review. I've read a forum thread in the Kodi forum about multiprocessing (I did not save it). It was mentioned that one should not use real multiprocessing because it causes problems on different platforums. As I understood, multiprocessing.dummy is not real multiprocessing but only multithreading, so I hope that this will not cause any problems. But let's see what @Lunatixz thinks about that approach.

plugin.video.telebasel/resources/lib/telebasel.py
goggle: Sorry, I did not see your reply. How do I know how to trigger the fallback? Will these distributions with lightweight python variants not include the `multiprocessing` module, so `import multiprocessing.dummy` will throw an `ImportError`? Or how do I know when to use the fallback?

enen92:
Yeah I think the issue I had with multiprocessing was on my rpi running libreelec. Are you able to test it there @goggle ?

goggle:
@enen92 I have a RPI with a LibreElec nightly build, but I did not test it there. I will implement a fallback mode and push it onto this PR.

goggle:
Ok, I've added a fallback mode for systems without the `multiprocessing` module. If everything is alright, let me know and I will squash the commits.

enen92:
@goggle all fine. Please squash for merging (and thanks for adjusting)

PR #2119, opened 2018-10-20
[plugin.video.tyt] 3.0.3

lottaboost:
Fixed language file.  Was using en_us, instead of the default.


Rechi:
@lottaboost check is failing because you didn't bump the version.

PR #2133, opened 2018-11-03
[plugin.video.frikanalen] v0.0.2

petterreinholdtsen:
### Description
A new version of the plugin.video.frikanalen already in jarvis.  News include HD stream
and API lookup of todays videos.

Replaces pull request #2132.
### Checklist:


plugin.video.frikanalen/addon.py
```diff
@@ -0,0 +1,81 @@
...
+from xbmcplugin import endOfDirectory
+
+import frikanalen
+
+plugin = routing.Plugin()
+
+@plugin.route('/')
+def root():
+    items = [
+        (plugin.url_for(live), ListItem("Direkte"), True),
```
enen92: would be nice if you could turn those listitems labels into localized strings

petterreinholdtsen: I would love to make them translatable, but have not been able to figure out how yet.

enen92: See https://kodi.wiki/view/Language_support

PR #2141, opened 2018-11-06
[plugin.video.frikanalen] v0.0.3

petterreinholdtsen:
### Description
- Added HD stream option.
- Added category listing of published videos.
- Show video duration in schedule and category lists.
- Updated to not crash with new API date format.
- Added free text video search option.
### Checklist:

### DIFF
--- a/plugin.video.frikanalen/addon.py
+++ b/plugin.video.frikanalen/addon.py
@@ -17,7 +17,10 @@
 
 
 import routing
+import urllib
+import xbmc
 import xbmcplugin
+from xbmcaddon import Addon
 from xbmcgui import ListItem
 from xbmcplugin import addDirectoryItem
 from xbmcplugin import addDirectoryItems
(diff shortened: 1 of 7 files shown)


plugin.video.frikanalen/addon.py
```diff
@@ -27,55 +30,108 @@
...
 @plugin.route('/live')
 def live():
-    addon_handle = plugin.handle
-
-    xbmcplugin.setContent(addon_handle, 'videos')
+    xbmcplugin.setContent(plugin.handle, 'videos')
 
-    li = ListItem('Frikanalen-sending (SD)', iconImage='DefaultVideo.png')
+    li = ListItem(_('Frikanalen akkurat nå (SD)'), iconImage='DefaultVideo.png')
     li.setProperty('IsPlayable', 'true')
```
enen92: Iconimage in deprecated please use setart

plugin.video.frikanalen/addon.py
```diff
@@ -27,55 +30,108 @@
...
 
     info = {'mediatype': 'video'}
     # info['plot'] = ''
     li.setInfo('video', info)
 
-    addDirectoryItem(handle=addon_handle, url=frikanalen.stream_url(), listitem=li)
+    addDirectoryItem(handle=plugin.handle, url=frikanalen.stream_url(), listitem=li)
 
-    li = ListItem('Frikanalen-sending (HD)', iconImage='DefaultVideo.png')
+    li = ListItem(_('Frikanalen akkurat nå (HD)'), iconImage='DefaultVideo.png')
```
enen92: It would be nice if you could turn this into localized strings

plugin.video.frikanalen/frikanalen.py
```diff
@@ -90,7 +142,34 @@ def whats_on():
     for item in program:
         t = item.starttime
         if t.hour == now.hour:
-            print("> [{:d}:{:02d}] {:s} {:s}".format(t.hour, t.minute, item.video.name, item.duration))
+            print("> [{:d}:{:02d}] {:s} {:f}".format(t.hour, t.minute, item.video.name, item.duration))
```
enen92: Please use xbmc.log instead of print statements

plugin.video.frikanalen/addon.py
petterreinholdtsen: This is a localized string.

plugin.video.frikanalen/frikanalen.py
petterreinholdtsen: This print code is so far only used by the self test, which I run outside Kodi.  It is a goal to keep the frikanalen.py code workout without any xbmc dependencies to allow the API access to be tested without having the xbmc* libraries installed.  The print lines will be changed/removed when this code is called in the addon code path.

plugin.video.frikanalen/addon.py
petterreinholdtsen: OK.  I'll change this and update the pull request branch.

PR #2172, opened 2018-11-28
[plugin.video.science.friday] 3.0.2

willforde:
### Description

Migrate to package setup for smaller entry point.
Some small tweaks as well.

### Checklist:


plugin.video.science.friday/resources/lib/main.py
```diff
@@ -0,0 +1,133 @@
...
+        url = "https://www.sciencefriday.com/wp-admin/admin-ajax.php?action=get_results&paged=%(next)s&" \
+              "sfid=%(sfid)s&post_types=%(ctype)s&_sft_topic=%(topic)s" % \
+              {"sfid": sfid, "ctype": ctype, "topic": topic, "next": page_count}
+    else:
+        url = "https://www.sciencefriday.com/wp-admin/admin-ajax.php?action=get_results&paged=%(next)s&" \
+              "sfid=%(sfid)s&post_types=%(ctype)s" % \
+              {"sfid": sfid, "ctype": ctype, "next": page_count}
+
+    # Fetch & parse HTML Source
+    ishd = bool(plugin.setting.get_int("video_quality", addon_id="script.module.youtube.dl"))
```
Rechi: `script.module.youtube.dl` is missing in the addon.xml, but code looks like it's not optional.

willforde:
Oops. It would have worked anyway since youtube.dl is a dependency of CodeQuick. But your right it should still be added sense I'm directly accessing it.

Rechi:
I've seen that CodeQuick has it as dependency, but the dependency might get removed in a future version.

PR #2174, opened 2018-11-28
[plugin.video.mvmix] 2.0.0

JinRonin:
Music Video add-on to play local and online music video randomly.
Its (currently) using imvdb and youtube as video sources and last.fm to search for similar artists. 

Been using this for a couple of years and moving it now from another (kodinerds) repo to the official one.


plugin.video.mvmix/resources/lib/artist_player.py
```diff
@@ -0,0 +1,103 @@
+# -*- coding: utf-8 -*-
+
+import random
+import sys
+import xbmc
+import xbmcgui
+import xbmcplugin
+
+class mvmixArtistPlayer(xbmc.Player):
```
enen92: Plugins should not use xbmc.Player this way. Why don't you just create a playlist?

JinRonin: Sorry didnt know. I use it to create a playlist to automatically add new videos everytime a new video starts.
As this is core functionality, feel free to do whatever with this pr. 
Im not able to rethink/rework this anytime soon.

PR #2214, opened 2018-12-29
[plugin.video.dumpert] 1.1.7

skipmodea1:
### Description
v1.1.7 (2018-12-28)
- added stream videos
### Checklist:


Rechi:
I'm fine with allowing it into isengard, as one can see it is as update (already existed in gotham).

The version in gotham branch should be marked broken.
So this (isengard versions) may need a higher version number to have a free one for the broken version.

PS: @skipmodea1 please look up how to ammend and squash commits, as opening new PRs instead of updating an existing one is annoying.

PR #2226, opened 2019-01-10
[plugin.program.remote.control.browser] 1.4.2

chadparry:
### Description
This small update fixes some warnings and an error in the previous version.

### Checklist:
- [x] Each add-on submission should be a single commit with using the following style: [plugin.video.foo] 1.0.0


plugin.program.remote.control.browser/browse.py
```diff
@@ -17,31 +17,35 @@
...
+    logger.debug('Missing Python package: pylirc')
     pylirc = None
 
 
-logger = logging.getLogger('remotecontrolbrowser')
-logger.addHandler(logging.StreamHandler())
-logger.setLevel(logging.INFO)
-
-
 VOLUME_MIN = 0L
```
Rechi: The `L` literal for numbers is incompatible with python3 (https://www.python.org/dev/peps/pep-0237/). As the following variables are only used where the value is between `-1` and `101` it's not really necessary to define them as long. So it would be great if you can adjust the code.
Please tell me if you plan to do it in this PR or in the next one.

```suggestion
VOLUME_MIN = 0
```

plugin.program.remote.control.browser/browse.py
```diff
@@ -17,31 +17,35 @@
...
 VOLUME_MAX = 100L
```
Rechi:
```suggestion
VOLUME_MAX = 100
```

plugin.program.remote.control.browser/browse.py
```diff
@@ -17,31 +17,35 @@
...
 DEFAULT_VOLUME = 50L
```
Rechi:
```suggestion
DEFAULT_VOLUME = 50
```

plugin.program.remote.control.browser/plugin.py
```diff
@@ -29,15 +29,13 @@
...
 except ImportError:
+    xbmc.log('Missing Python package: pulsectl', xbmc.LOGDEBUG)
     pulsectl = None
-try:
-    import pyctl
-except ImportError:
-    pyctl = None
 
 
 DEFAULT_VOLUME = 50L
```
Rechi:
```suggestion
DEFAULT_VOLUME = 50
```

chadparry:
As long as you don't mind, I'll make the Python 3 changes in a different pull request. There are a lot more extensive changes that I will need to do besides just removing the `L` literal. For example, I'll have to use the new `urllib` and `http` libraries, remove the `long` function, stop using the `u` Unicode literal, and change character encodings.

Rechi:
I know there are more code changes required, to really make it python 3 compatible, but one has to start somewhere.

PR #2338, opened 2019-03-15
[plugin.video.rsiplaytv] v1.5.0

goggle:
### Description
As the previously merged plugins `plugin.video.srfplaytv` and `plugin.video.rtsplaytv`, this is the initial published version of `plugin.video.rsiplaytv`, which covers the Italian part of the Swiss national TV media library.

### Checklist:

Additional information :

I have a question about the updating process. Let's say I update the library `script.module.srgssr` to version 1.5.1. Should I then also update the plugins like `plugin.video.srfplaytv`, which use `script.module.srgssr` as a dependency, even if I don't need to change anything in the plugin itself (except the version number of the plugin)? Is there a "best practice" or is this completely up to me?


Rechi:
As mentioned at https://kodi.wiki/view/Submitting_Add-ons#Allowed_submissions, new additions are only allowed for Leia.

\
The version attribute in the import tags actually means minimum version. If installing plugin (and dependencies) from repo, the users gets the last available version, regardless which is specified in addon.xml (as long as it is equal or higher).

There is no strict rule how to handle it from our side, but if you need new functions from the dependency library, version should be updated. In that case addon-checker will tell the order of deploying. Otherwise we might deploy the plugin update before the dependency library update and it will be broken for some time.

goggle:
@Rechi Thanks for the detailed answer!

Ok, then I will move this PR to the Leia branch. Is there anything that needs to be changed for this (e.g. version of `xbmc.python`?)

plugin.video.rsiplaytv/addon.xml
```diff
@@ -0,0 +1,38 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<addon id="plugin.video.rsiplaytv" name="RSI Play TV" version="1.5.0" provider-name="Alexander Seiler">
+  <requires>
+    <import addon="xbmc.python" version="2.25.0"/>
```
Rechi:
```suggestion
    <import addon="xbmc.python" version="2.26.0"/>
```

plugin.video.rsiplaytv/resources/lib/rsiplaytv.py
```diff
@@ -0,0 +1,169 @@
...
+    from urllib import unquote_plus
+
+from kodi_six import xbmc, xbmcaddon, xbmcplugin
+import srgssr
+
+ADDON_ID = 'plugin.video.rsiplaytv'
+REAL_SETTINGS = xbmcaddon.Addon(id=ADDON_ID)
+ADDON_NAME = REAL_SETTINGS.getAddonInfo('name')
+ADDON_VERSION = REAL_SETTINGS.getAddonInfo('version')
+DEBUG = (REAL_SETTINGS.getSetting('Enable_Debugging') == 'true')
```
Rechi:
```suggestion
DEBUG = REAL_SETTINGS.getSettingBool('Enable_Debugging')
```

plugin.video.rsiplaytv/resources/lib/rsiplaytv.py
```diff
@@ -0,0 +1,169 @@
...
+    try:
+        page = unquote_plus(params['page'])
+    except Exception:
+        page = None
+
+    log('Mode: ' + str(mode))
+    log('URL : ' + str(url))
+    log('Name: ' + str(name))
+    log('Page Hash: ' + str(page_hash))
+    log('Page: ' + str(page))
```
Rechi: combine this into one log line

goggle:
@Rechi Thanks for the review. I will include the suggested changes in my next PR of `plugin.video.rsiplaytv` in the Leia branch.

PR #2368, opened 2019-04-15
[plugin.video.dazn] 1.3.2

JinRonin:
added youth protection pin (mainly for german users)

### DIFF
--- a/plugin.video.dazn/README.md
+++ b/plugin.video.dazn/README.md
@@ -1,5 +1,5 @@
 # DAZN Video Plugin for Kodi (plugin.video.dazn)
- 
+
 ## Disclaimer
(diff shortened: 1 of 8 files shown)


dc95336:
What's the protection pin be for

JinRonin:
Some content is protected in germany (pg-18) and you have to verify your age with dazn in order to watch it.
The update allows you to enter the pin you create.

dc95336:
Ok thank you

Rechi:
@JinRonin this add-on isn't Python 3 compatible, but all add-ons in Leia branch and higher have to be Python 3 compatible (see https://kodi.tv/article/attention-addon-developers-migration-python-3). Seems like the add-on was accepted before this rule was enforced. So please make your add-on Python 3 compatible.

PR #2380, opened 2019-04-27
[plugin.audio.soundcloud] 3.0.0

jaylinski:
### Description
This is an incomplete rewrite of the original (and broken) plugin developed by bromix.

#### Features
* Search music
* Charts
* Discover new music
* Backwards compatible `/play/` API
* Python 3 compatibility

#### Known issues
* Login is not yet implemented
* Settings are not backwards compatible (settings from v2 will be lost)

### About this PR
* This is my first plugin, so I guess I made some errors. Please review carefully.
* Settings are not backwards compatible. I guess I could make some settings backwards compatible (like the search result length), but I think it's not worth the time.
* Plugin is newly built from ground up, because of Python 3 compatibility
* I'm using xbmc.python version 2.25.0 deliberately, because some users still use v17 (although they won't get this update)

### Source
https://github.com/jaylinski/kodi-addon-soundcloud

### Checklist:


plugin.audio.soundcloud/README.md
```diff
@@ -0,0 +1,77 @@
...
+
+This [Kodi](https://github.com/xbmc/xbmc) Add-on provides a minimal interface for SoundCloud.
+
+## Features
+
+* Search
+* Discover new music
+* Play tracks, albums and playlists
+
+## Installation
```
jaylinski: This has to be changed if the plugin is accepted. I guess a link to https://kodi.wiki/view/Add-on:SoundCloud should suffice.

plugin.audio.soundcloud/resources/lib/kodi/cache.py
```diff
@@ -0,0 +1,31 @@
+import os
+import time
+
+
+class Cache:
```
jaylinski: If you can point me to a good and maintained Kodi caching module, I'll use it instead of this module.

plugin.audio.soundcloud/resources/lib/soundcloud/api_v2.py
```diff
@@ -0,0 +1,238 @@
...
+from resources.lib.models.user import User
+from resources.lib.soundcloud.api_collection import ApiCollection
+from resources.lib.soundcloud.api_interface import ApiInterface
+
+
+class ApiV2(ApiInterface):
+    """This class uses the unofficial API used by the SoundCloud website."""
+
+    api_host = "https://api-v2.soundcloud.com"
+    api_client_id = "KT6UCHXC9iNnI8wn4UUfwMSlAPe4Z8zx"
```
jaylinski: This API key changes regularly, I don't know how long it stays valid. But the key can be overridden via settings.

The key should be automatically extracted from the SoundCloud website. Otherwise the plugin has to be updated as soon as the key expires.

plugin.audio.soundcloud/resources/plugin.py
```diff
@@ -0,0 +1,146 @@
...
+import xbmc
+import xbmcaddon
+import xbmcgui
+import xbmcplugin
+
+addon = xbmcaddon.Addon()
+addon_id = addon.getAddonInfo("id")
+addon_base = "plugin://" + addon_id
+addon_profile_path = xbmc.translatePath(addon.getAddonInfo('profile')).decode('utf-8')
+addon_temp_path = xbmc.translatePath('special://temp/%s' % addon_id).decode('utf-8')
```
jaylinski: According to the guidelines, there has to be a good reason for **not** writing data to the userdata directory.

I saw that the twitch plugin also writes cache data to the temp folder, so I'm also using the temp dir for storing the cache data.

PR #2381, opened 2019-04-28
[plugin.audio.soundcloud] 3.0.0

jaylinski:
### Description
This is an incomplete rewrite of the original (and broken) plugin developed by bromix.

> Since this is an update, the target is Krypton. If this is regarded as a new plugin, I will target it to Leia.

#### Features
* Search music
* Charts
* Discover new music
* Backwards compatible `/play/` API
* Python 3 compatibility

#### Known issues
* Login is not yet implemented
* Settings are not backwards compatible (settings from v2 will be lost)

### About this PR
* This is my first plugin, so I guess I made some errors. Please review carefully.
* Settings are not backwards compatible. I guess I could make some settings backwards compatible (like the search result length), but I think it's not worth the time.
* Plugin is newly built from ground up, because of Python 3 compatibility

### Source
https://github.com/jaylinski/kodi-addon-soundcloud

### Checklist:


plugin.audio.soundcloud/README.md
```diff
@@ -0,0 +1,77 @@
...
+* Search
+* Discover new music
+* Play tracks, albums and playlists
+
+## Installation
+
+* [Download the latest release](https://github.com/jaylinski/kodi-addon-soundcloud/releases) (`plugin.audio.soundcloud.zip`)
+* Copy the zip file to your Kodi system
+* Open Kodi, go to Add-ons and select "Install from zip file"
+* Select the file `plugin.audio.soundcloud.zip`
```
jaylinski: This has to be changed if the plugin is accepted. I guess a link to https://kodi.wiki/view/Add-on:SoundCloud should suffice.

plugin.audio.soundcloud/resources/lib/kodi/cache.py
```diff
@@ -0,0 +1,31 @@
+import os
+import time
+
+
+class Cache:
```
jaylinski: If you can point me to a good and maintained Kodi caching module, I'll use it instead of this module.

plugin.audio.soundcloud/addon.xml
```diff
@@ -0,0 +1,65 @@
...
+[upd] framework
+
+1.2.1 (2015-05-17)
+1.2.0 (2015-05-09)
+1.1.2 (2015-02-27)
+1.1.1 (2015-01-25)
+1.1.0 (2015-01-14)
+1.0.2 (2014-11-29)
+1.0.1 (2014-11-27)
+1.0.0 (2014-10-29)
```
jaylinski: Changelog for v1 has been truncated because of the 1500 char limit.

plugin.audio.soundcloud/resources/lib/soundcloud/api_v2.py
```diff
@@ -0,0 +1,238 @@
...
+from resources.lib.models.user import User
+from resources.lib.soundcloud.api_collection import ApiCollection
+from resources.lib.soundcloud.api_interface import ApiInterface
+
+
+class ApiV2(ApiInterface):
+    """This class uses the unofficial API used by the SoundCloud website."""
+
+    api_host = "https://api-v2.soundcloud.com"
+    api_client_id = "KT6UCHXC9iNnI8wn4UUfwMSlAPe4Z8zx"
```
jaylinski: This API key changes regularly, I don't know how long it stays valid. But the key can be overridden via settings.

The key should be automatically extracted from the SoundCloud website. Otherwise the plugin has to be updated as soon as the key expires.

plugin.audio.soundcloud/resources/plugin.py
```diff
@@ -0,0 +1,146 @@
...
+import xbmc
+import xbmcaddon
+import xbmcgui
+import xbmcplugin
+
+addon = xbmcaddon.Addon()
+addon_id = addon.getAddonInfo("id")
+addon_base = "plugin://" + addon_id
+addon_profile_path = xbmc.translatePath(addon.getAddonInfo('profile')).decode('utf-8')
+addon_temp_path = xbmc.translatePath('special://temp/%s' % addon_id).decode('utf-8')
```
jaylinski: According to the guidelines, there has to be a good reason for not writing data to the userdata directory.

I saw that the twitch plugin also writes cache data to the temp folder, so I'm also using the temp dir for storing the cache data.

plugin.audio.soundcloud/resources/lib/soundcloud/api_v2.py
jaylinski: Is there a hook, so I can extract the key on every Kodi start-up?

jaylinski: I think this can be done with a service (https://kodi.wiki/view/Service_add-ons). At least Twitch is doing it like this.

jaylinski: I implemented it in https://github.com/jaylinski/kodi-addon-soundcloud/pull/9 and will push it upstream as soon as I receive the first feedback.

plugin.audio.soundcloud/README.md
tadly: I can't so you're good :)

plugin.audio.soundcloud/resources/lib/kodi/cache.py
tadly: I can't so you're good :)

plugin.audio.soundcloud/resources/lib/soundcloud/api_v2.py
tadly: Only did some quick googeling and to me it seems as v2 would require you to implement oauth to officially get authentication related keys (which would include refresh-tokens etc.)

This would be a little bit more involved but also be a cleaner solution.

If I'm missing something please let me know as I really didn't do much research on this.

P.s. I raised this internally to see if other are okay with this solution (for now/indefinitely)

plugin.audio.soundcloud/resources/plugin.py
tadly: As per the add-on rules wiki entry:
> add-ons should store all their data in their own subfolder inside the addon_data directory.

So if you could change that to the profile dir, that would be great :)
`xbmc.translatePath(addon.getAddonInfo('profile'))` (Not sure if `translatePath` is required)

tadly:
First of, sorry it took so long.
Second, congraz on this clean and well structure code. :)
Third, I myself couldn't find any major concerns from what you already commented

plugin.audio.soundcloud/addon.xml
```diff
@@ -0,0 +1,65 @@
...
+        <summary lang="de_DE">Soundcloud ist eine Online-Audio-Vertriebsplattform.</summary>
+		<summary lang="en_GB">SoundCloud is an online audio distribution platform.</summary>
+        <language></language>
+        <platform>all</platform>
+        <license>MIT</license>
+        <forum>https://forum.kodi.tv/showthread.php?tid=206635</forum>
+        <website>https://soundcloud.com</website>
+        <source>https://github.com/jaylinski/kodi-addon-soundcloud</source>
+        <news>3.0.0 (2019-0x-xx)
+[new] complete rewrite
```
enen92: news tag is only for the latest version, if you need to keep a changelog along with the code use the changelog.txt (although deprecated)

plugin.audio.soundcloud/resources/lib/soundcloud/api_v2.py
enen92: You can create a service extension point to extract the new key on kodi on specific time-intervals

plugin.audio.soundcloud/resources/plugin.py
enen92: Yes, please cache on the userdata folder and manage it in the addon. /temp data won't have a reference to the running profile

plugin.audio.soundcloud/resources/lib/soundcloud/api_v2.py
enen92: Another option is to decorate the methods that require the key to a request (prior to the method execution) which will grab the key and cache it. If the last request don't exceed a specific threshold just use the cached one. If the request fails with auth-related http status codes just grab it again, and so on

tadly: Even better, make `api_client_id` a property which handles the caching.
Then other functions don't need decorators and it's a easy drop-in replacement.

Something like:
```python
...
@property
def api_client_id(self):
  if <should_fetch_id_again_because_enough_time_passed>:
    # fetch key
    self._api_client_id = <new id>
    # write to cache

  if not self._api_client_id:
    # self._api_client_id = <read from cache>

  return self._api_client_id
...
```

plugin.audio.soundcloud/resources/plugin.py
```diff
@@ -0,0 +1,146 @@
...
+import sys
+import xbmc
+import xbmcaddon
+import xbmcgui
+import xbmcplugin
+
+addon = xbmcaddon.Addon()
+addon_id = addon.getAddonInfo("id")
+addon_base = "plugin://" + addon_id
+addon_profile_path = xbmc.translatePath(addon.getAddonInfo('profile')).decode('utf-8')
```
romanvm: `.decode('utf-8')` is going to break in Python 3. I'd recommend to use kodi-six: https://github.com/romanvm/kodi.six

jaylinski: Good to know! Did I miss this in the docs?
I got this from the wiki, which should probably be updated: https://kodi.wiki/view/Add-on_rules

jaylinski: This page should also contain some infos on Python 3 compatibility: https://kodi.wiki/view/Add-on_unicode_paths

plugin.audio.soundcloud/resources/lib/soundcloud/api_v2.py
jaylinski: @enen92 @tadly thanks for your feedback!
I will try your proposed approaches in the future (https://github.com/jaylinski/kodi-addon-soundcloud/issues/11). If it's ok for you, I'll stick with the service-approach for now.

Regarding the cleaner approach with the official oauth implementation: SoundCloud locked the app-registration about a year ago: http://soundcloud.com/you/apps/new
So it's impossible to use the official API if you didn't register an app a year ago... :man_shrugging:

plugin.audio.soundcloud/resources/plugin.py
jaylinski: I fixed it. The plugin now uses a `cache` subdirectory inside the profile folder:
```

jaylinski:
@tadly @enen92 Thanks for your time and your feedback!

I pushed the updated code.

I'm still unsure how I should handle the UTF-8 issue mentioned by @romanvm ...

romanvm:
@jaylinski Since you are trying to make your code compatible with Python 3, you can either use `kodi-six` or to do conditional decoding. Something like this:
```python
from future.utils import PY2

addon_profile_path = xbmc.translatePath(addon.getAddonInfo('profile'))
if PY2:
    addon_profile_path = addon_profile_path.decode('utf-8')

jaylinski:
@romanvm Thanks, that was exactly what I was looking for! I tested it and pushed it.

plugin.audio.soundcloud/addon.xml
jaylinski: Fixed. I also replaced the temporary date with a real date.

plugin.audio.soundcloud/resources/lib/soundcloud/api_v2.py
tadly: @jaylinski ah. Well that's a bummer but good to know :)

Rechi:
@jaylinski this add-on isn't Python 3 compatible. Future updates have to be compatible with Python 3.

jaylinski:
> @jaylinski this add-on isn't Python 3 compatible. Future updates have to be compatible with Python 3.
> Have a look at https://kodi.tv/article/attention-addon-developers-migration-python-3 for more information.

@Rechi Thanks for the info. I will test the plugin with a Kodi Python 3 build. PR will follow.

PR #2384, opened 2019-05-03
[plugin.audio.vrt.radio] v0.1.0

dagwieers:
### Description
This is the first release of this addon, it includes all the public streams from the Flemish broadcast organization VRT.

Originally this was a patch to the existing **plugin.video.vrt.nu** addon, but it was decided to split off the Radio functionality to its own addon (so it appears with a different icon and name).

Although the version is quite low, it is already full-featured :-)

### Checklist:


Rechi:
New add-ons are only allowed for Leia, see https://kodi.wiki/view/Submitting_Add-ons#Allowed_submissions.

dagwieers:
@Rechi That is unfortunate. Fixed in PR #2384

PR #2389, opened 2019-05-07
[plugin.audio.vrt.radio] v0.1.0

dagwieers:
### Description
This is the first release of this addon, it includes all the public streams from the Flemish broadcast organization VRT.

Originally this was a patch to the existing plugin.video.vrt.nu addon, but it was decided to split off the Radio functionality to its own addon (so it appears with a different icon and name).

Although the version is quite low, it is already full-featured :-)

This fixes #2384

### Checklist:


plugin.audio.vrt.radio/addon.xml
```diff
@@ -0,0 +1,37 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<addon id="plugin.audio.vrt.radio" name="VRT Radio" version="0.1.0" provider-name="dagwieers">
+    <requires>
+        <!-- import addon="resource.images.studios.coloured" version="0.0.15"/ -->
+        <import addon="resource.images.studios.white" version="0.0.21"/>
+        <import addon="script.module.routing" version="0.2.0"/>
+        <import addon="xbmc.python" version="2.25.0"/>
```
Rechi: WARN: For leia it is advised to set xbmc.python version to 2.26.0
```suggestion
        <import addon="xbmc.python" version="2.26.0"/>
```

dagwieers: But the plugin also works on krypton. So why artificially limit to Leia here, if it will only ship to Leia? It means I have to remember changing this for this repository ever release.

PR #2439, opened 2019-07-10
[plugin.video.vrt.nu] 2.1.0

dagwieers:
### Description
- A small bugfix wrt. roaming tokens
- Now includes YouTube support and one missing radio channel
- A new API call to start a program based on start-time

### Checklist:

### DIFF
--- a/plugin.video.vrt.nu/resources/lib/kodiwrapper.py
+++ b/plugin.video.vrt.nu/resources/lib/kodiwrapper.py
@@ -179,7 +179,7 @@ class KodiWrapper:
             is_folder = bool(not title_item.is_playable and title_item.path)
             is_playable = bool(title_item.is_playable and title_item.path)
 
-            list_item = xbmcgui.ListItem(label=title_item.title)
+            list_item = xbmcgui.ListItem(label=title_item.title, offscreen=True)
             list_item.setProperty(key='IsPlayable', value='true' if is_playable else 'false')
 
             # FIXME: The setIsFolder is new in Kodi18, so we cannot use it just yet.
@@ -191,6 +191,10 @@ class KodiWrapper:
             if title_item.info_dict:
                 # type is one of: video, music, pictures, game
                 list_item.setInfo(type='video', infoLabels=title_item.info_dict)
+
+            if title_item.stream_dict:
+                # type is one of: video, audio, subtitle
+                list_item.addStreamInfo('video', title_item.stream_dict)
 
             if title_item.context_menu:
                 list_item.addContextMenuItems(title_item.context_menu)
@@ -207,7 +211,7 @@ class KodiWrapper:
     def play(self, video):
         ''' Create a virtual directory listing to play its only item '''
         import xbmcgui
-        play_item = xbmcgui.ListItem(path=video.stream_url)
+        play_item = xbmcgui.ListItem(path=video.stream_url, offscreen=True)
         play_item.setProperty('inputstream.adaptive.max_bandwidth', str(self.get_max_bandwidth() * 1000))
         play_item.setProperty('network.bandwidth', str(self.get_max_bandwidth() * 1000))
         if video.stream_url is not None and video.use_inputstream_adaptive:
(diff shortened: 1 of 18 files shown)


plugin.video.vrt.nu/resources/lib/kodiwrapper.py
```diff
@@ -229,7 +233,14 @@ def play(self, video):
...
 
         self.log_notice('Play: %s' % unquote(video.stream_url), 'Info')
-        xbmcplugin.setResolvedUrl(self._handle, bool(video.stream_url), listitem=play_item)
+
+        # To support video playback directly from RunPlugin() we need to use xbmc.Player().play instead of
+        # setResolvedUrl that only works with PlayMedia() or with internal playable menu items
+        if self._handle != -1:
+            xbmcplugin.setResolvedUrl(self._handle, bool(video.stream_url), listitem=play_item)
+        else:
+            xbmc.Player().play(item=video.stream_url, listitem=play_item)
```
basrieter: It is not allowed to call the `xbmc.Player().play()` method from Python. Add-ons should always use the `xbmcplugin.setResolvedUrl()` method.

dagwieers:
cc @mediaminister

plugin.video.vrt.nu/resources/lib/kodiwrapper.py
dagwieers: cc @mediaminister

mediaminister: @basrieter Why? We need to use `RunPlugin` to remotely start a stream from a plugin url. In that case the add-on handle has a value of `-1`. But `setResolvedUrl` doesn't work then and complains with `WARNING: Attempt to use invalid handle -1`.
I didn't find a way to successfully call `setResolvedUrl` from `RunPlugin`. Any ideas?

basrieter: @Lunatixz can provide you the details. But I assume (which is the mother of all ....) that if you are using the xbmc.Player().play() because of the -1 handle, you are not creating a plugin, but rather a script?

mediaminister: I wanted to add an external interface to play a stream based on EPG data and this should be part of the existing  VRT NU plugin, I just want ` RunPlugin("plugin://plugin.video.vrt.nu/play/airdate/een/2019-07-15T13:00:00")` to directly play a stream. But I doubt whether this is possible if I'm not allowed to use `xbmc.Player().play()`. Not sure how to fix this. cc @michaelarnauts

michaelarnauts: I'm okay with just PlayMedia support, but that kills the ability to display a new directory listing in case we ever want to (maybe when there are multiple results or something).

I'm using this to play past programs from the PVR guide with this plugin: https://github.com/michaelarnauts/context.pvr.justmissed. I hope to integrate with various other add-ons that allow to play recently aired programs.

dagwieers: @basrieter Is this a hard requirement? Because it clearly adds functionality that is useful in a very minimalistic way. This was added deliberately and not as an oversight (i.e. because we did not know what we were doing). And I don't want to delay the release of this version over functionality that is not used by our existing users.

mediaminister: @dagwieers: It's okay for me to remove xbmc.Player().play() because the airdate function doesn't use directory listings and by design returns a single video item to `setResolvedUrl`. We can use `PlayMedia` instead of `RunPlugin` without losing any functionality:
PlayMedia("plugin://plugin.video.vrt.nu/play/airdate/een/2019-07-15T13:00:00")

dagwieers: @mediaminister OK, I'll remove it for the v2.1.0 release and decide about the below in our codebase.

> but that kills the ability to display a new directory listing in case we ever want to (maybe when there are multiple results or something).

dagwieers:
Thanks !

PR #2440, opened 2019-07-10
[plugin.video.sandmann] v1.0.2

sorax:
### Description
Watch the current episode of "Sandmann" (a show for children with a short bedtime story each day).

### Checklist:


sorax:
Updated to v1.0.1 (python3 compatible)

(thanks for your hint with the force push)

plugin.video.sandmann/sandmann.py
```diff
@@ -0,0 +1,75 @@
...
+
+try:
+    from urllib.request import urlopen
+except ImportError:
+    from urllib2 import urlopen
+
+try:
+    import json
+except ImportError:
+    import simplejson as json
```
Rechi: Replace this block with `import json`. Python 2.7 and higher always (guaranteed for Kodi Leia) has the json module.

sorax: I updated the mentioned part.
Thank you so much for your assistance!!

PR #2444, opened 2019-07-13
[plugin.video.embycon] 1.9.18

faush01:
### Description
July Release
A lot of small fixes and compatibility changes for the latest version of the Emby 4.2 Server.

I hope this is the last Python 2.7 release!
I am waiting for nightly Kodi 19 builds with Python 3 to switch my attention to that, hopefully it is soon.

### Checklist:

Additional information :

### DIFF
--- a/plugin.video.embycon/resources/lib/websocket.py
+++ b/plugin.video.embycon/resources/lib/websocket.py
@@ -21,6 +21,7 @@ Copyright (C) 2010 Hiroki Ohtani(liris)
 
 
 import socket
+from base64 import b64encode
 
 try:
     import ssl
@@ -186,7 +187,10 @@ def _parse_url(url):
     if parsed.query:
         resource += "?" + parsed.query
 
-    return (hostname, port, resource, is_secure)
+    user_name = parsed.username
+    user_password = parsed.password
+
+    return (hostname, port, resource, is_secure, user_name, user_password)
 
 
 def create_connection(url, timeout=None, **options):
@@ -455,7 +459,8 @@ class WebSocket(object):
                  the custom HTTP headers are added.
 
         """
-        hostname, port, resource, is_secure = _parse_url(url)
+        hostname, port, resource, is_secure, user_name, user_password = _parse_url(url)
+
         # TODO: we need to support proxy
         self.sock.connect((hostname, port))
         if is_secure:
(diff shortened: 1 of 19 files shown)


faush01:
The above are not errors for Python 2.7

basrieter:
It has to be compatible with both Py2 and Py3: 

> After the release of Kodi 18 (Leia) only addons (new and/or updates) that are compatible with both Python 2 and 3 or Python 3-only will be accepted to the repositories

See: https://forum.kodi.tv/showthread.php?tid=327746

faush01:
I am happy to start work on a switch to Python 3 but I am not going to be working on Python 2 and 3 compatibility in the same code. Too many headaches.

take the above example for instance:
iteritems() and items() do different things on Python2, iteritems() does what items() does on Python3, I am not going to be spending time on testing how this effect performance on multiple platforms especially when this code base will never touch Python3.

As I said I am happy to start work on Python 3 when there are nightly builds from main that have Python3 in them but until then I will be sticking with Kodi 18 Python 2.

faush01:
I will close this and submit a deprecation version

plugin.video.embycon/resources/lib/websocket.py
```diff
@@ -894,8 +906,13 @@ def run_forever(self, sockopt=None, sslopt=None, ping_interval=0):
                     self._callback(self.on_message, data)                    
                     
                 except Exception as e:
-                    #print str(e.args[0])
-                    if "timed out" not in e.args[0]:
+                    found_timeout = False
+                    for arg in e.args:
+                        if isinstance(arg, str):
+                            print str(arg)
```
Rechi: The following change would have been enough to get CI green.
```suggestion
                            print(str(arg))
```

faush01: But it is always something, I come to dread having to try to get the addon updated, it is not worth the stress and issues.

faush01: Nothing against the project, it is just not for me, I am going to go back to a private repo, it is just so much easier to manage and update.

faush01:
@Rechi 
I think I just need to understand what is going on.
Why is there a requirement for both Pthon 2 and 3 compatibility, yes you are switching from 2 to 3 some time in the future but the current Kodi 18 is Python 2 and that is what the addon is targeting. I dont see all the fuss about concurrent Python 2 and 3 compatibility, when Kodi 19 has Python 3 then people can start testing and working out how to best support both version or just make a clean cut to Python 3.
This should be left up to the addon devs, it should not be forced on them.
We don't even have Kodi 19 nightly build with Python 3 to test on yet.

Its not just the print() lines, the iterator changes were also failing in the build, the problem is the changes requested would potentially fix a Python 2 and 3 dual comparability requirements but the items() and iteritems() does different things in Python 2.
Also it looks like the verification you are using is only checking the changed lines, so this will not catch all compatibility issues.

As I have said, once Kodi 19 has nightly builds that have Python 3 then I will start looking at it but until then there is no point. As a plugin dev I feel that should be my call.

Please help me understand.

Rechi:
As your add-on is in the krypton branch you are currently not forced to be completely Python 3 compatible (although it would be nice if you are already).

There are a a few checks (old Python 2 except, exec, ne, print, raise, repr and tuple_params syntax) which fail the build as the new Python 3 syntax also works in Python 2. The others, like items/iteritems, are only reported as information and don't fail the check as they are not always needed.

faush01:
@Rechi 
I remember fixing all the old Python 2 checks a few versions ago but though I was fine now until Kodi 19.

But the items/iteritems was breaking the build, I will create a new PR and try again.

Rechi:
Trust me, items/iteritems didn't break the build. Only the print line caused the failure, which was previously commented out.

faush01:
Ok thanks for the help, I will try again.

Issue #2409, opened 2019-06-08
[Broken] Preview Networks


KarellenX:
**[Broken] Preview Networks**

### Add-on details:

- **General**
  - Add-on name: **plugin.video.previewnetworks**
  - Add-on ID: 
  - Version number: **2.3.5**
  - Kodi/repository version: **All**

https://kodi.tv/addon/plugins-video-add-ons/preview-networks
https://kodi.wiki/view/Add-on:Preview_Networks
https://forum.kodi.tv/showthread.php?tid=70179
https://previewnetworks-xbmc-plugin.googlecode.com/svn/trunk/plugin.video.previewnetworks/
https://code.google.com/archive/p/previewnetworks-xbmc-plugin/

- **Code location**
  - URL: **Does not exist**
  - Revision/tag: **N/A**
  - Branch: **N/A**

### Description of changes:
Add-on is available through official repo, but the content source site no longer exists. The add-on installs without error, but no content is available.

PR #2466, opened 2019-07-27
[plugin.video.previewnetworks] broken

enen92:
Requested by @KarellenX in https://github.com/xbmc/repo-plugins/issues/2409


KarellenX:
Thank you @enen92

PR #2471, opened 2019-08-03
[plugin.video.rivedila7] 3.1.1

luivit:
### Checklist:

Additional information :
- New version, compatible from kodi 17


enen92:
@luivit your add-on is already on the Jarvis branch. Any reason for submitting it to krypton?

Testato:
The new 3.x version need the inputstream adaptive
<import addon="inputstream.adaptive"/>

It is not present on kodi16

enen92:
Ok, we don't accept new addons in the krypton branch but I guess this can be considered an update to an already existing addon. Please mark the jarvis version as broken if it doesn't work anymore. Also keep in mind that from Leia and upper kodi versions, addons have to be compatible with python3

Rechi:
@luivit this add-on isn't Python 3 compatible. Future updates have to be compatible with Python 3.

PR #2482, opened 2019-08-14
[plugin.video.vrt.nu] 2.2.0

dagwieers:
### Description
The new version includes some cosmetic changes, it has breadcrumbs support, various fixes to credentials and tokens, a new metadata engine and new artwork.

### Checklist:


plugin.video.vrt.nu/resources/lib/kodiwrapper.py
```diff
@@ -100,6 +100,13 @@ def has_socks():
     return has_socks.installed
 
 
+class SafeDict(dict):
```
basrieter: Why not use a standard `dict.get(key, fallback)`?

plugin.video.vrt.nu/resources/lib/kodiwrapper.py
```diff
@@ -437,61 +486,83 @@ def get_path(self, path):
...
-        import xbmcvfs
+        from xbmcvfs import delete
         self.log_notice("Delete file '%s'." % path, 'Debug')
-        return xbmcvfs.delete(path)
-
-    def md5(self, path):
-        ''' Return an MD5 checksum of a file '''
+        return delete(path)
+
+    def delete_cached_thumbnail(self, url):
```
basrieter: You are not allowed to change/delete outside your own directories. You should use the proper Kodi API using `Textures.GetTextures` and `Textures.RemoveTexture` (See https://kodi.wiki/view/JSON-RPC_API/v8). 

I understand the reason behind the cache removal, but you are not updating the texture database like this. You could also try to update the texture by adding a random URL parameters? I guess it should be seen as a new thumb?  Not 100% sure.

dagwieers: cc @mediaminister

mediaminister: Okay, I rewrote `delete_cached_thumbnail` function using Kodi JSON-RPC API to comply with the rules: https://github.com/pietje666/plugin.video.vrt.nu/pull/443
@dagwieers: Can you add this to this release?

dagwieers: I think we are going for a v2.0.1 release instead. Since v2.0.0 was already cut from wood.
(And handed out)

plugin.video.vrt.nu/resources/lib/kodiwrapper.py
dagwieers: This is being used when doing pyformat in localized formatted strings.

If an unknown `{foobar}` is encountered, we do not want it to be reason for failure, but instead show `{foobar}` in the resulting string.

PR #2538, opened 2019-09-30
[plugin.video.vrt.nu] 2.2.3

dagwieers:
### Description
- Integrate new inputstreamhelper
- Support user paths with unicode characters
- Manage My favorites
- Improve breadcrumbs support
- Various translation fixes

### Checklist:

### DIFF
--- a/plugin.video.vrt.nu/resources/lib/kodiwrapper.py
+++ b/plugin.video.vrt.nu/resources/lib/kodiwrapper.py
@@ -16,7 +16,7 @@ try:  # Python 3
 except ImportError:  # Python 2
     from urllib2 import unquote
 
-sort_methods = dict(
+SORT_METHODS = dict(
     # date=xbmcplugin.SORT_METHOD_DATE,
     dateadded=xbmcplugin.SORT_METHOD_DATEADDED,
     duration=xbmcplugin.SORT_METHOD_DURATION,
@@ -30,11 +30,18 @@ sort_methods = dict(
     unsorted=xbmcplugin.SORT_METHOD_UNSORTED,
 )
 
-log_levels = dict(
+LOG_LEVELS = dict(
     Quiet=0,
     Info=1,
     Verbose=2,
     Debug=3,
+)
+
+XBMC_LOG_LEVELS = dict(
+    Quiet=xbmc.LOGNONE,
+    Info=xbmc.LOGNOTICE,
+    Verbose=xbmc.LOGINFO,
+    Debug=xbmc.LOGDEBUG,
 )
 
 WEEKDAY_LONG = {
@@ -90,14 +97,15 @@ MONTH_SHORT = {
 
 def has_socks():
     ''' Test if socks is installed, and remember this information '''
-    if not hasattr(has_socks, 'installed'):
-        try:
-            import socks  # noqa: F401; pylint: disable=unused-variable,unused-import
-            has_socks.installed = True
-        except ImportError:
-            has_socks.installed = False
-            return None  # Detect if this is the first run
-    return has_socks.installed
+    if hasattr(has_socks, 'installed'):
+        return has_socks.installed
+    try:
+        import socks  # noqa: F401; pylint: disable=unused-variable,unused-import
+        has_socks.installed = True
+        return True
+    except ImportError:
+        has_socks.installed = False
+        return None  # Detect if this is the first run
 
 
 class SafeDict(dict):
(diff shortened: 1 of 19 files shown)


plugin.video.vrt.nu/resources/lib/kodiwrapper.py
```diff
@@ -635,49 +623,53 @@ def refresh_caches(self, cache_file=None):
...
-            xbmc.log(msg=from_unicode(message), level=xbmc.LOGNOTICE)
+        cur_log_level = LOG_LEVELS.get(log_level, 0)
+        if not self._debug_logging and 1 < cur_log_level <= self._max_log_level:
+            # If Debug Logging is not enabled, Kodi filters everything up to NOTICE out
+            log_level = 'Info'
+        if kwargs:
+            import string
+            message = string.Formatter().vformat(message, (), SafeDict(**kwargs))
+        message = '[{addon}] {message}'.format(addon=self._addon_id, message=message)
+        xbmc.log(msg=from_unicode(message), level=XBMC_LOG_LEVELS.get(log_level))
```
basrieter: Logging to the Kodi log is only allowed with the `DEBUG` level. See https://kodi.wiki/view/Add-on_rules:

> addons should use the xbmc.log() method to write to the logfile and use the debug logging level only.

plugin.video.vrt.nu/resources/lib/kodiwrapper.py
```diff
@@ -635,49 +623,53 @@ def refresh_caches(self, cache_file=None):
...
 
-    def log_error(self, message, log_level='Info'):
+    def log_error(self, message, **kwargs):
         ''' Log error messages to Kodi '''
-        message = '[%s] %s' % (self._addon_id, message)
+        if kwargs:
+            import string
+            message = string.Formatter().vformat(message, (), SafeDict(**kwargs))
+        message = '[{addon}] {message}'.format(addon=self._addon_id, message=message)
         xbmc.log(msg=from_unicode(message), level=xbmc.LOGERROR)
```
basrieter: Logging to the Kodi log is only allowed with the `DEBUG` level. See https://kodi.wiki/view/Add-on_rules:

> addons should use the xbmc.log() method to write to the logfile and use the debug logging level only.

dagwieers: This is the case. But we make an exception for testing/development.

plugin.video.vrt.nu/resources/lib/kodiwrapper.py
dagwieers: This is the case. But we make an exception for testing/development.

plugin.video.vrt.nu/resources/lib/kodiwrapper.py
```diff
@@ -396,10 +389,13 @@ def get_max_bandwidth(self):
     def get_proxies(self):
         ''' Return a usable proxies dictionary from Kodi proxy settings '''
         usehttpproxy = self.get_global_setting('network.usehttpproxy')
-        if usehttpproxy is False:
+        if usehttpproxy is not True:
```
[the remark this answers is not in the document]
dagwieers: Not necessarily, it could be a string, or None. And in that case we consider it False, rather than True.

plugin.video.vrt.nu/resources/lib/kodiwrapper.py
basrieter: But in that case, you do write to the Kodi log with other levels?

ronie:
hmm, why are you using the xbmc.python.module extension point? that doesn't seem right...
https://github.com/dagwieers/repo-plugins/blob/662adfb5c33a613dfeeb814700a818b04f97054b/plugin.video.vrt.nu/addon.xml#L19

plugin.video.vrt.nu/resources/lib/kodiwrapper.py
DaVukovic: This is not resolved yet

> This is the case. But we make an exception for testing/development.

Testing and development should be done before an add-on is submitted. Beside that, I'm very curious who the "we" is in that case. As "we" (in meaning of TeamKodi) haven't made any exception I could think of. So please follow our add-on rules and change what others have requested.

Thanks in advance

dagwieers: @DaVukovic When I said *"We make an exception"* I meant that from the settings.xml we (the add-on developers) allow the user to change the log level so that we can have a detailed add-on logging without a detailed Kodi logging. This helps in troubleshooting add-on specific issues.

dagwieers:
@ronie This is to ensure that for the add-on *resources/lib/* is part of the `PYTHONPATH`. Otherwise to python *resources/* and *lib/* are considered python packages and need to be included when importing our own python classes.

I would much rather have Kodi to include *resources/lib/* by default in an add-on's `PYTHONPATH` but that is currently not the case.

plugin.video.vrt.nu/resources/lib/kodiwrapper.py
dagwieers: @basrieter Incorrect. By default nothing is logged (this depends on `max_log_level` in *settings.xml*), but when `debug_logging` is `True`, we force `xbmc.LOGDEBUG` for everything and we force the maximum logging in the add-on.

basrieter: So then I can't follow your code. If the `debug_logging` is false, the log level is set to `Info`, which logs in the kodi.log with level `Notice`? Which is against the rules?

dagwieers: Well, there does seem to be a problem, but it is not related to the fact we check for debug_logging, it is related to the fact we don't filter out based on max_log_level.

The whole point of this change is NOT to log with LOGNOTICE by default.

PR #2544, opened 2019-10-03
[plugin.video.psvue] 2019.10.3

eracknaphobia:
### Description

- Add EPG service (uses IPTV Simple PVR)
- DVR resumetime default
- Code Cleanup
### Checklist:

Additional information :

### DIFF
--- a/plugin.video.psvue/main.py
+++ b/plugin.video.psvue/main.py
@@ -1,9 +1,11 @@
 from resources.lib.ps_vue import *
+import xbmcgui
 
 params = get_params()
 url = None
 name = None
 mode = None
+vod = 'null'
 airing_id = 'null'
 channel_id = 'null'
 program_id = 'null'
@@ -23,6 +25,8 @@ if 'title' in params:
     title = params["title"]
 if 'plot' in params:
     plot = params["plot"]
+if 'vod' in params:
+    vod = params["vod"]
 if 'icon' in params:
     icon = params["icon"]
 if 'mode' in params:
@@ -114,6 +118,18 @@ elif mode == 850:
 elif mode == 900:
     get_stream(url, airing_id, channel_id, program_id, series_id, tms_id, title, plot, icon)
 
+elif mode == 950:
+    if vod != airing_id:
+        choice = xbmcgui.Dialog().yesno("Where would you like to watch this episode?","Click an item below to choose your preference", nolabel='ON DEMAND', yeslabel='DVR')
+        if choice:
+            get_stream(url, airing_id, channel_id, program_id, series_id, tms_id, title, plot, icon)
+        else:
+            airing_id = vod
+            url = SHOW_URL + vod
+            get_stream(url, airing_id, channel_id, program_id, series_id, tms_id, title, plot, icon)
+    else:
+        get_stream(url, airing_id, channel_id, program_id, series_id, tms_id, title, plot, icon)
+
 elif mode == 997:
     epg_service = xbmcaddon.Addon('service.psvue.epg')
     epg_file_path = xbmc.translatePath(epg_service.getSetting(id='location'))
@@ -121,7 +137,6 @@ elif mode == 997:
     epg_toggle_off = '{"jsonrpc": "2.0", "method": "Addons.SetAddonEnabled", ' \
                      '"params": {"addonid": "service.psvue.epg", "enabled": false}, "id": 1}'
     xbmc.executeJSONRPC(epg_toggle_off)
-
     xbmcvfs.delete(os.path.join(epg_file_path, 'epg.db'))
     xbmcvfs.delete(os.path.join(epg_file_path, 'epg.xml'))
     xbmcvfs.delete(os.path.join(epg_file_path, 'playlist.m3u'))
--- a/plugin.video.psvue/resources/lib/globals.py
+++ b/plugin.video.psvue/resources/lib/globals.py
@@ -1,0 +1,153 @@
+import cookielib
+from datetime import datetime, timedelta
+import os, sys
+import threading
+import time, math
+import requests
+import random
+import urllib
+import xbmc, xbmcgui, xbmcaddon, xbmcvfs, xbmcplugin
+import _strptime
+
+
+if not xbmc.getCondVisibility('System.HasAddon(pvr.iptvsimple)'):
+    dialog = xbmcgui.Dialog()
+    dialog.notification('PS Vue EPG', 'Please enable PVR IPTV Simple Client', xbmcgui.NOTIFICATION_INFO, 5000, False)
+    sys.exit()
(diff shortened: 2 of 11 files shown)


plugin.video.psvue/main.py
```diff
@@ -114,14 +118,25 @@
 elif mode == 900:
     get_stream(url, airing_id, channel_id, program_id, series_id, tms_id, title, plot, icon)
 
+elif mode == 950:
+    if vod != airing_id:
+        choice = xbmcgui.Dialog().yesno("Where would you like to watch this episode?","Click an item below to choose your preference", nolabel='ON DEMAND', yeslabel='DVR')
```
Lunatixz: Best to move string to localized po.

plugin.video.psvue/resources/lib/globals.py
```diff
@@ -0,0 +1,153 @@
...
+import requests
+import random
+import urllib
+import xbmc, xbmcgui, xbmcaddon, xbmcvfs, xbmcplugin
+import _strptime
+
+
+if not xbmc.getCondVisibility('System.HasAddon(pvr.iptvsimple)'):
+    dialog = xbmcgui.Dialog()
+    dialog.notification('PS Vue EPG', 'Please enable PVR IPTV Simple Client', xbmcgui.NOTIFICATION_INFO, 5000, False)
```
Lunatixz: Same

plugin.video.psvue/resources/lib/globals.py
```diff
@@ -0,0 +1,153 @@
...
+    try:
+        date = datetime.strptime(str(string), date_format)
+    except TypeError:
+        date = datetime(*(time.strptime(str(string), date_format)[0:6]))
+    return date
+
+
+def check_iptv_setting(id, value):
+    if IPTV_SIMPLE_ADDON.getSetting(id) != value:
+        IPTV_SIMPLE_ADDON.setSetting(id=id, value=value)
```
Lunatixz: Can't set another plugin's settings, you could set up a button to opensettings().

plugin.video.psvue/resources/lib/guideservice.py
```diff
@@ -0,0 +1,179 @@
...
+            self.guide_thread_2 = threading.Thread(name='GuideThread 2',
+                                                   target=self.short_guide(channel_ids[:third]))
+            self.guide_thread_3 = threading.Thread(name='GuideThread 3',
+                                                   target=self.short_guide(channel_ids[third:third + third]))
+            self.guide_thread_4 = threading.Thread(name='GuideThread 4',
+                                                   target=self.short_guide(channel_ids[third + third:]))
+
+            xbmc.log('BuildGuide: before loop')
+            thread_alive = True
+            while thread_alive:
```
Lunatixz: Please use `while not xbmc.abortRequested() and thread_alive:`

plugin.video.psvue/resources/lib/guideservice.py
```diff
@@ -0,0 +1,179 @@
...
+
+                if self.monitor.waitForAbort(5):
+                    break
+
+            xbmc.log('BuildGuide: after loop')
+            self.db.clean_db_epg()
+            self.db.build_epg_xml()
+            if first_time_thru:
+                xbmc.log('BuildGuide: First time thru, toggling IPTV restart')
+                pvr_toggle_off = '{"jsonrpc": "2.0", "method": "Addons.SetAddonEnabled", ' \
```
Lunatixz: Not sure if this is allowed without first prompting the user? @enen92 input?

plugin.video.psvue/main.py
eracknaphobia: Will do.

plugin.video.psvue/resources/lib/globals.py
eracknaphobia: Normally I would strongly agree with this rule but in this case I don't think it should apply. First this is a "opt-in" option and I'd argue that I am using IPTV Simple Client the way it is intended, by supplying it with a playlist & xmltv path. Nothing malicious or overstepping, I am simply automating a step that I believe most users wouldn't be able to or care to do on their own. The idea of popping a message and telling them how to do it instead of just doing it and knowing it is done correctly seems silly to me. As the former introduces user error that will inevitably fall back on the dev having to lend support to.

plugin.video.psvue/resources/lib/guideservice.py
enen92: Yes please add a yes/no before the call

plugin.video.psvue/resources/lib/globals.py
Lunatixz: I don't believe the repository has ever made an exception when it comes to writing to the settings of another plugin. IMO the whole IPTV simple thing is a bit hackish. Generating M3Us and XMLTV files are fine, manipulating Kodis api to shoehorn support seems problematic.

eracknaphobia: image
@enen92  @Rechi 
I'd like to request access to write to IPTV Simple settings in order to allow users to use PS Vue through the built-in Kodi EPG. This is a "opt-in" option.

Lunatixz: @eracknaphobia I don't think this situation applies; that statement applies to addon folders and writing files to them. ie Cache modules, etc. However,  it doesn't hurt to revisit the issue.

enen92: @eracknaphobia would you consider a yes/no before changing the setting? 
I'm fine with it as it is though.

eracknaphobia: @enen92 Yes, I have no problem doing that.

enen92: I'd also prefer that the code is not so generic and that it identifies clearly what settings you're trying to manipulate. I douby you need such a generic method.
what exactly do you need to do?

eracknaphobia: @enen92 
Basically I'm setting the option for playlist and epg location to local and setting the paths to the files in the ps vue addon_data folder. I'm also setting the logos to be pulled remotely and to prefer the logo urls in the m3u file.

This section should make it clearer as to what values are set.
https://github.com/xbmc/repo-plugins/blob/b9d2b3b1d36aebaa9b1288d590cc2b191d48c019/plugin.video.psvue/resources/lib/globals.py#L148-L153
https://github.com/xbmc/repo-plugins/blob/b9d2b3b1d36aebaa9b1288d590cc2b191d48c019/plugin.video.psvue/resources/lib/database.py#L140

eracknaphobia:
Closing because PS Vue is going offline permanently come January 2020. No reason to pursue this further.

PR #2579, opened 2019-10-31
[plugin.video.intergalacticfm] 2.1.1

dromer:
### Description
Version 2.1.1
Mtv url fix and some images improved

Version 2.1.0
Some channels added and some removed


plugin.video.intergalacticfm/README.md
```diff
@@ -21,6 +21,11 @@ Stepts to take to install from the Kodi main repository are:
 
 Instructions for installing Kodi can be found here: https://kodi.tv/download
 
+#### Android
+Please consider installing [Kodi](https://f-droid.org/en/packages/org.xbmc.kodi/)
```
Rechi: I wouldn't recommend installing Kodi from F-Droid. It doesn't provide the most recent stable release.

PanderMusubi: Normally, F-Droid automatically builds when there is a new tagged release availabe.

Rechi: see https://github.com/xbmc/xbmc/releases, https://play.google.com/store/apps/details?id=org.xbmc.kodi and https://f-droid.org/en/packages/org.xbmc.kodi/

dromer: Hmmm, seems they have AutoUpdateMode turned off indeed: https://gitlab.com/fdroid/fdroiddata/blob/master/metadata/org.xbmc.kodi.yml#L133

If you think we should remove any links for android users we can omit it from the readme.

PanderMusubi: Fine by me if that needs to be removed from README.md (also the audio plugin then).

Rechi: Yes, in instructions for how to get/intstall Kodi please only link to sites and stores controlled by Team Kodi.

PR #2583, opened 2019-11-04
[plugin.video.zoneminder] 2.0.1

petegallagher:
### Description
Update ZoneMinder plugin to allow users to modify the cgi-bin path. The RedHat version of ZoneMinder uses a different path from other distros so we need to make an option for this.

### Checklist:


Rechi:
@petegallagher this add-on isn't Python 3 compatible. Future updates have to be compatible with Python 3.

PR #2628, opened 2019-12-08
[plugin.video.sandmann] v1.1.0

sorax:
### Description
- **add**: reload functionality (can be configured in the addon-settings)
- **fix**: images not loading

### Checklist:

### DIFF
--- a/plugin.video.sandmann/addon.py
+++ b/plugin.video.sandmann/addon.py
@@ -41,10 +41,11 @@ interval = addon.getSettingInt("interval")
 title, description, thumbnail_image, fanart_image, stream = getEpisodeData(
     episodes_url, quality)
 
-li_episode = xbmcgui.ListItem(label=title, thumbnailImage=thumbnail_image)
-li_episode.setProperty("fanart_image", fanart_image)
+li_episode = xbmcgui.ListItem(label=title)
+li_episode.setArt({'thumb': thumbnail_image, 'fanart': fanart_image})
 li_episode.setInfo(type="Video", infoLabels={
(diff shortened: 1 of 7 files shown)


plugin.video.sandmann/addon.py
```diff
@@ -16,25 +16,53 @@
...
+quality = addon.getSettingInt("quality")
+update = addon.getSettingInt("update")
+interval = addon.getSettingInt("interval")
+
+title, description, thumbnail_image, fanart_image, stream = getEpisodeData(
     episodes_url, quality)
 
-li = xbmcgui.ListItem(label=title, thumbnailImage=thumbnail_image)
-li.setProperty("fanart_image", fanart_image)
+li_episode = xbmcgui.ListItem(label=title, thumbnailImage=thumbnail_image)
```
enen92: Please remove thumbnailImage kwarg and use the setArt method on the ListItem object:

```python
li_episode.setArt({'thumb':thumbnail_image, 'fanart':fanart_image})
```

plugin.video.sandmann/addon.py
```diff
@@ -16,25 +16,53 @@
...
+li_episode.setProperty("fanart_image", fanart_image)
```
enen92: Drop please

plugin.video.sandmann/addon.py
sorax: okay, I'll do this in the next version.

plugin.video.sandmann/addon.py
sorax: okay

PR #2648, opened 2019-12-23
CHG: disable Plugin

raptor2101:
### Description
simply disable plugin
### Checklist:


plugin.audio.podcatcher/addon.xml
```diff
@@ -31,5 +31,6 @@
     <website>https://github.com/raptor2101/PodCatcher</website>
     <source>https://github.com/raptor2101/PodCatcher</source>
     <email>raptor2101@gmx.de</email>
+    <broken>For updates of PodCatcher, please upgrade to krypton or above</broken>
```
enen92: a version bump is required.
Please use also the same format for the PR title: [plugin] broken

PR #2662, opened 2019-12-29
[plugin.audio.radio_de] 2.4.0

enen92:
v2.4.0 (29/12/2019)

- [new] Use radio API v2
- [fix] Remove dead code
- [new] Add radio.pt
- [new] Page listings
- [new] Better quality logos
- [new] Hide fanart setting

### DIFF
--- a/plugin.audio.radio_de/resources/lib/api.py
+++ b/plugin.audio.radio_de/resources/lib/api.py
@@ -38,14 +38,11 @@ class RadioApiError(Exception):
 class RadioApi():
 
     MAIN_URLS = {
-        'english': 'http://rad.io/info',
-        'german': 'http://radio.de/info',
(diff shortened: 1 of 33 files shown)


plugin.audio.radio_de/resources/lib/api.py
```diff
@@ -213,28 +302,41 @@ def __urlopen(self, url):
...
-            )
-            genre = station.get('genresAndTopics') or ','.join(
-                station.get('genres', []) + station.get('topics', []),
+                station.get('logo300x300') or
+                station.get('logo175x175') or
+                station.get('logo100x100') or
+                station.get('logo44x44')
             )
+
+            try:
```
basrieter: Could you not do `genre = [g.get('value', d) for g in station.get('genres')]`

Same for the descriptions and names

PR #2682, opened 2020-01-18
[plugin.video.vrt.nu] v2.3.2

dagwieers:
### Description
- Fix issues related to categories scraping and caching (@mediaminister)
- Various improvements to TV guide (@dagwieers)

This is another bugfix release. Without the bugfix some of our users will have issues with the categories listing in 2 days (it expires after 7 days, since our last release).

### Checklist:

### DIFF
--- a/plugin.video.vrt.nu/README.md
+++ b/plugin.video.vrt.nu/README.md
@@ -50,7 +50,11 @@ leave a message at [our Facebook page](https://facebook.com/kodivrtnu/).
 ![](./resources/media/screenshot04.jpg)
 
 ## Releases
-### v2.3.1 (2020-01-13)
+### v2.3.2 (2020-01-18)
+- Fix issues related to categories scraping and caching (@mediaminister)
+- Various improvements to TV guide (@dagwieers)
+
+### v2.3.1 (2020-01-14)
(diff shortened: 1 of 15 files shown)


dagwieers:
The error does not relate to our add-on, since it does not use **urllib3** or **requests**.
Something must be wrong with the build infrastructure.

enen92:
@dagwieers xsd schema for addon.xml is defined to a max of 1500 chars: `<xs:maxLength xmlns:xs="http://www.w3.org/2001/XMLSchema" value="1500" /> `. Please try to keep it to the last release only (it is not a full changelog)

dagwieers:
@enen92 We trim the news-section, but prefer to keep a few releases in there. I would have cleaned it up if Travis would have given me a sensible error :-) Thanks for your help!

enen92:
The reason for the char limit is because, since news is part of addon.xml, it'll be available on the xml.gz manifest kodi downloads from the mirrors periodically. So if everyone abuses on the size of the news element, it causes harm to the infrastructure
Cheers, merged

dagwieers:
Sure, I have no problem with the 1500 char limit. And we are not adding the complete changelog.

PR #2701, opened 2020-02-04
[plugin.audio.mixcloud] v.2.4.2

jackyNIX:
### Description
fixed local resolver
added german language

### Checklist:


jackyNIX:
I know about the python 3 change, but didn't had the time to really get into it. 
This commit is just a minor change. Only 1 line of code, but necessary to keep it working. So if this could be merged as is, that would be cool.
Next one will be python 3.

PR #2730, opened 2020-03-01
[plugin.video.vtm.go] 1.0.0

michaelarnauts:
### Description
*plugin.video.vtm.go* is a Kodi add-on for watching live video streams and video-on-demand content available on the VTM GO platform. 

It has been out for a few months already, and I was waiting for the inputstream.adaptive 2.4.3 release for Kodi 18, since this addon depends on that version.

### Checklist:

Additional information :


plugin.video.vtm.go/CHANGELOG.md
```diff
@@ -0,0 +1,221 @@
+# Changelog
```
dagwieers: In previous reviews the Kodi devs have requested to *not* include changelogs and use the addon.xml news section for including recent changelog information.

michaelarnauts: I've removed the `CHANGELOG.md` file.

plugin.video.vtm.go/resources/lib/modules/channels.py
```diff
@@ -0,0 +1,164 @@
...
+            channel_info = next((c for c in channel_infos if c.key == key), None)
+
+            if channel_info:
+                # Lookup the high resolution logo based on the channel name
+                icon = '{path}/resources/logos/{logo}-white.png'.format(path=self._kodi.get_addon_path(), logo=channel.get('logo'))
+                fanart = '{path}/resources/logos/{logo}.png'.format(path=self._kodi.get_addon_path(), logo=channel.get('logo'))
+
+                context_menu = [(
+                    self._kodi.localize(30052, channel=channel.get('label')),  # Watch live {channel}
+                    'XBMC.PlayMedia(%s)' %
```
enen92: `PlayMedia` is enough. `XBMC.PlayMedia` is just legacy stuff, you can easily tell how old it is :)

plugin.video.vtm.go/resources/lib/modules/channels.py
```diff
@@ -0,0 +1,164 @@
...
+                    self._kodi.url_for('play', category='channels', item=channel_info.channel_id)
+                ), (
+                    self._kodi.localize(30053, channel=channel.get('label')),  # TV Guide for {channel}
+                    'XBMC.Container.Update(%s)' %
```
enen92: same

plugin.video.vtm.go/resources/lib/modules/channels.py
```diff
@@ -0,0 +1,164 @@
...
+                    self._kodi.url_for('show_tvguide_channel', channel=channel.get('epg'))
+                )]
+                if self._kodi.get_setting_as_bool('metadata_update'):
+                    context_menu.append(
+                        (
+                            self._kodi.localize(30055, channel=channel.get('label')),  # Catalog for {channel}
+                            'XBMC.Container.Update(%s)' %
```
enen92: same

plugin.video.vtm.go/resources/lib/modules/menu.py
```diff
@@ -0,0 +1,294 @@
...
+        prop_dict = {}
+
+        #
+        # Movie
+        #
+        if isinstance(item, Movie):
+            if item.my_list:
+                context_menu = [(
+                    self._kodi.localize(30101),  # Remove from My List
+                    'XBMC.Container.Update(%s)' %
```
enen92: same

plugin.video.vtm.go/resources/lib/modules/menu.py
```diff
@@ -0,0 +1,294 @@
...
+                    self._kodi.url_for('mylist_del', video_type=self._vtm_go.CONTENT_TYPE_MOVIE, content_id=item.movie_id)
+                )]
+            else:
+                context_menu = [(
+                    self._kodi.localize(30100),  # Add to My List
+                    'XBMC.Container.Update(%s)' %
```
enen92: same

plugin.video.vtm.go/resources/lib/modules/menu.py
```diff
@@ -0,0 +1,294 @@
...
+                             is_playable=True)
+
+        #
+        # Program
+        #
+        if isinstance(item, Program):
+            if item.my_list:
+                context_menu = [(
+                    self._kodi.localize(30101),  # Remove from My List
+                    'XBMC.Container.Update(%s)' %
```
enen92: same

plugin.video.vtm.go/resources/lib/modules/menu.py
```diff
@@ -0,0 +1,294 @@
...
+                    self._kodi.url_for('mylist_del', video_type=self._vtm_go.CONTENT_TYPE_PROGRAM, content_id=item.program_id)
+                )]
+            else:
+                context_menu = [(
+                    self._kodi.localize(30100),  # Add to My List
+                    'XBMC.Container.Update(%s)' %
```
enen92: same

plugin.video.vtm.go/resources/lib/modules/menu.py
```diff
@@ -0,0 +1,294 @@
...
+
+        #
+        # Episode
+        #
+        if isinstance(item, Episode):
+            context_menu = []
+            if item.program_id:
+                context_menu = [(
+                    self._kodi.localize(30102),  # Go to Program
+                    'XBMC.Container.Update(%s)' %
```
enen92: same

plugin.video.vtm.go/resources/lib/plugin.py
```diff
@@ -0,0 +1,207 @@
...
+def play_epg_datetime(channel, timestamp):
+    """ Play a program based on the channel and the timestamp when it was aired """
+    from resources.lib.modules.tvguide import TvGuide
+    TvGuide(kodi).play_epg_datetime(channel, timestamp)
+
+
+@routing.route('/play/catalog/<category>/<item>/<channel>')
+def play_or_live(category, item, channel):
+    """ Ask to play the requested item or switch to the live channel """
+    from resources.lib.modules.player import Player
```
enen92: Plugins should not call player.Play directly, please resolve the listitem path and use `xbmcplugin.setResolvedUrl`. This way any time you play a stream, a new entry will be saved on the kodi database.

plugin.video.vtm.go/resources/lib/modules/tvguide.py
```diff
@@ -0,0 +1,129 @@
...
+            self._kodi.show_notification(message=str(ex))
+            self._kodi.end_of_directory()
+            return
+
+        listing = []
+        for broadcast in epg.broadcasts:
+            if broadcast.playable_type == 'episodes':
+                context_menu = [(
+                    self._kodi.localize(30102),  # Go to Program
+                    'XBMC.Container.Update(%s)' %
```
enen92: another

plugin.video.vtm.go/resources/lib/plugin.py
enen92: This is a blocker unfortunately.

plugin.video.vtm.go/resources/lib/plugin.py
```diff
@@ -0,0 +1,207 @@
...
+    Player(kodi).play_or_live(category, item, channel)
+
+
+@routing.route('/play/catalog/<category>/<item>')
+def play(category, item):
+    """ Play the requested item """
+    from resources.lib.modules.player import Player
+    Player(kodi).play(category, item)
```
enen92: same story. You can still use the player class to wait for playback to start and feed it subtitiles afterwards but calling play directly has unwanted consequences. Furthermore, you loose features of the plugin addon type such as watch states, resume, etc.

plugin.video.vtm.go/resources/lib/vtmgo/vtmgostream.py
```diff
@@ -0,0 +1,473 @@
...
+        info = json.loads(response.text)
+        return info
+
+    def _anvato_get_stream_info(self, anvato_info, stream_info):
+        """ Get the stream info from anvato.
+        :type anvato_info: dict
+        :type stream_info: dict
+        :rtype dict
+        """
+        url = 'https://tkx.apis.anvato.net/rest/v2/mcp/video/{video}'.format(**anvato_info)
```
enen92: Not familiar with this. @basrieter is probably the guy to clarify.

plugin.video.vtm.go/service.py
```diff
@@ -0,0 +1,8 @@
+# -*- coding: utf-8 -*-
+""" Service entry point """
+
+from __future__ import absolute_import, division, unicode_literals
+
+from resources.lib import service
+
+service.run()
```
enen92: since the addon already includes a service maybe you can move the player object there and avoid sleeping after requesting playback.

plugin.video.vtm.go/resources/lib/vtmgo/vtmgostream.py
michaelarnauts: This just fills in the `video` placeholder from the `anvato_info` object.

This would be the same as 
```
        url = 'https://tkx.apis.anvato.net/rest/v2/mcp/video/{video}'.format(video=anvato_info['video'])
```

plugin.video.vtm.go/resources/lib/plugin.py
michaelarnauts: This `Player` class is my own class, not to be confused with Kodi's player. It just groups some related code to prepare an item to play. In the end, I'm using `setResolvedUrl`.

plugin.video.vtm.go/resources/lib/plugin.py
michaelarnauts: As the other remark, this Player().play() is from `resources.lib.modules.player`, and this is just a class to do metadata lookup and in the end uses `setResolvedUrl`.

michaelarnauts:
I have removed the XBMC prefixes.

I will squash when everything is approved, this makes it easier to keep track of the requested changes.

michaelarnauts:
@enen92 I think the `Player` class caused some confusion. You can find the implementation here: https://github.com/xbmc/repo-plugins/pull/2730/files#diff-ca93b2330dd2d8684692efa005011a60

The `Player.play()` function is called to do metadata lookups, and then calls `self._kodi.play` from the KodiWrapper. There, the `Kodiwrapper.play()` function calls `xbmcplugin.setResolvedUrl()` at line 226 (https://github.com/xbmc/repo-plugins/pull/2730/files#diff-8623844f32f431b81d37cbcc03475c73R226)

plugin.video.vtm.go/resources/lib/vtmgo/vtmgostream.py
basrieter: Indeed.

plugin.video.vtm.go/resources/lib/plugin.py
basrieter: @enen92 The `Player` class is their own class and it passes their `KodiWrapper` object to the constructor. Their Player(kodi).play_or_live() then calls the `KodiWrapper.play()` which then calls `xbmcplugin.setResolvedUrl(self._handle, True, listitem=play_item)`. So I think they are good here. The code trail is very confusing, but they are not calling the `xbmc.Player().play()`.

plugin.video.vtm.go/resources/lib/plugin.py
basrieter: @enen92 Same here.

plugin.video.vtm.go/resources/lib/service.py
```diff
@@ -0,0 +1,81 @@
...
+
+    def _update_metadata(self):
+        """ Update the metadata for the listings. """
+        from resources.lib.modules.metadata import Metadata
+
+        # Clear outdated metadata
+        self.kodi.invalidate_cache(self.cache_expiry)
+
+        # Create progress indicator
+        progress = self.kodi.show_progress_background(message=self.kodi.localize(30715))
```
basrieter: This is something that I don't like. This pops a progress bar every now and then when the service is updating meta data. As this is a **background** activity, and hence, no user has to wait for, we don't need any kind of progress bars appearing.

plugin.video.vtm.go/resources/lib/kodiwrapper.py
```diff
@@ -0,0 +1,583 @@
...
+            play_item.setProperty('inputstream.adaptive.license_key', license_key)
+
+        if title_item.subtitles_path:
+            play_item.setSubtitles(title_item.subtitles_path)
+
+        # To support video playback directly from RunPlugin() we need to use xbmc.Player().play instead of
+        # setResolvedUrl that only works with PlayMedia() or with internal playable menu items
+        xbmcplugin.setResolvedUrl(self._handle, True, listitem=play_item)
+
+        if self.get_setting_as_bool('showsubtitles'):
```
basrieter: This won't work: if a stream is already playing (video or audio), the `xbmc.Player().isPlaying()` will be `True` and the subtitles will be show on that stream, not the one that will start playing. I had the exact same issue in Retrospect. You will also need to check that `xbmc.Player().isPlaying()` and that it is playing your requested url.

plugin.video.vtm.go/resources/lib/service.py
michaelarnauts: Okay, I agree!

plugin.video.vtm.go/resources/lib/kodiwrapper.py
michaelarnauts: Hmm, this makes sense. This code is based on how the VRT NU addon does this, maybe they also have the same issue then. I'll update this!

michaelarnauts:
@basrieter I've removed the background progress and I've also using a `xbmc.Player` based class to wait for the playback to have started before we enable the subtitles.

michaelarnauts:
> So the player stuff is OK I think. But the visible progress bar on background work needs to go in my opinion. A user should not care for background work, and thus not see a progress bar randomly appearing.
> 
> And the subtitle code will only work if no stream is playing while starting a new stream. If one is already active, the code will set the subtitles on the currently playing stream, not the new one.

Actually, the subtitles are added on the `ListItem` with `setSubtitles()`, so the subtitle itself should be added fine on the new stream, but I guess there could still be a race condition where the `showSubtitles(True)` could be executed while the old stream was still playing.

This change actually could also fix a possible race condition where we would send the UpNext notification to early, so this checking on the URL is a nice thing.

PR #2736, opened 2020-03-02
[plugin.video.mediathek] 0.10.0

raptor2101:
### Description
Migration to Python3. [rerelease]
### Checklist:


sarbes:
Hi.

I've noticed some issues with your add-on. Some of them are not major, but some of them are show stoppers. 

The more serious problems are:
- 3sat is broken. The new Mediathek uses the ZDF backend software, so the implementation should be easy.
- ARD is partially broken. Instead of listing items, the add-on tries to play them. This broken behavior results in a crash due to a bug in Kodi.  
- The Kika scraper hammers the server by doing a GET for each listitem. This is outright bad and unacceptable.
- Hardcoded categories and show names. This should be fetched at runtime.
- Hardcoded language strings.
- A lot of times, Kodi fails to play mp4 streams (might be a server issue, use HLS!).

Other problems are:
- Funky stream selection. 
  - Instead of choosing HLS streaming which every API provides, the add-on tries to match a bitrate defined in its setting with a very low default.
  - Choosing the highest quality level enables HLS streams (sometimes?), but without using inputstream.adaptive. This results in 30s loading times.
  - The quality selector prefers the audio description over all other language tracks (ZDF).
- No metadata, the listitems are pretty bare. Most APIs offer a thumbnail, fanart, plot, ...
- Messy menu structure. Instead of organizing each broadcaster menu in a similar fashion, each one is randomly thrown together.
- No or limited API usage. Every broadcaster has an API. It offers more information and it is far more stable than a regex based HTML scraper.
- The description mentions ORF and NDR, both are not implemented.

Sorry, but in its current state, the add-on has no chance of being merged. 

I'm the developer of a lot of add-ons for the German speaking part of the community, so this is my "home turf". I've written scrapers for all the websites you are accessing. To be honest, there are so many issues that a rewrite is advisable. 

I'd suggest to start over with an add-on for a currently not implemented website, like the ORF. I'm available at #kodi-dev at freenode if you want to chat.

raptor2101:
@sarbes: you are aware of, that this is only the PR for a migration to Py3. The Plugin itself is already in your repo.

If this can't be merged. Please merge #2735 ...

sarbes:
After some team feedback, we decided to close your PR and mark your add-on as broken for now.

Please don't take this personal. I encourage you to continue writing code for Kodi. If you happen to have any questions, feel free to hit me up on our IRC channel.

PR #2802, opened 2020-04-04
[plugin.video.curiositystream] 1.0.0

GMaxera:
Description

This add-on enables browsing and playing content from your [Curiositystream](https://curiositystream.com) account

Zip package for testing the add-on: 
[plugin.video.curiositystream.zip](https://github.com/xbmc/repo-plugins/files/4432504/plugin.video.curiositystream.zip)

Checklist:
- [x] My code follows the add-on rules and piracy stance of this project.
- [x] I have read the CONTRIBUTING document


plugin.video.curiositystream/main.py
```diff
@@ -0,0 +1,205 @@
+import functools
```
enen92, line 1: usually this file, as it is the entrypoint should be as small as possible. Just import the main codebase and run. The reason behind it is the fact that this is the only file that's not compiled to bitcode

plugin.video.curiositystream/main.py
```diff
@@ -0,0 +1,205 @@
...
+import sys
+from contextlib import contextmanager
+from urlparse import parse_qsl
```
enen92, line 4: not python3 compatible, you might want to make it py2 and py3 compatible to ease the migration and future submission to the matrix branch

plugin.video.curiositystream/main.py
GMaxera: I've reduced as much as possible 👍

plugin.video.curiositystream/main.py
GMaxera: I've used try/expect but not sure if it's the best practice in that cases. Please, let me know

enen92:
```python
import sys

PY3 = sys.version_info.major

if PY3:
    your py3 import logic
else:
   your py2 import logic
```

GMaxera: Thanks :-) Done 👍

PR #2850, opened 2020-04-25
plugin.video.themoviedb.helper 2.4.36

jurialmunkey:
### Description

Matrix version of TMDbHelper plugin.
The plugin is PY2/3 compatible so only difference against Leia version is LOGNOTICE now LOGINFO and python dep bumped to 3.0.0
Leia PR Here: https://github.com/xbmc/repo-plugins/pull/2849

### Checklist:

Additional information :


plugin.video.themoviedb.helper/addon.xml
```diff
@@ -0,0 +1,76 @@
...
+  <import addon="script.module.requests" version="2.9.1"/>
+  <import addon="script.module.pil" version="1.1.7"/>
+</requires>
+<extension point="xbmc.python.pluginsource" library="plugin.py">
+  <provides>video</provides>
+</extension>
+<extension point="xbmc.service" library="service.py" />
+<extension point="xbmc.python.script" library="script.py" />
+<extension point="kodi.context.item">
+  <menu id="kodi.core.main">
```
enen92, line 18: @basrieter I thought now we need to use args?

enen92: nevermind, those are different files for each menu item

plugin.video.themoviedb.helper/addon.xml
```diff
@@ -0,0 +1,76 @@
...
+    <menu>
+      <label>TMDbHelper Options</label>
+      <item library="context-play.py">
+        <label>Play Using</label>
```
enen92, line 22: I'm not 100% familiar with context menu items but can't those be localized?

plugin.video.themoviedb.helper/addon.xml
```diff
@@ -0,0 +1,76 @@
...
+        <visible>!String.IsEmpty(ListItem.Property(Container.Info)) + !String.IsEmpty(ListItem.Property(tmdb_id)) + [String.IsEqual(ListItem.dbtype,episode) | String.IsEqual(ListItem.dbtype,movie) | String.IsEqual(ListItem.dbtype,tvshow)]</visible>
+      </item>
+    </menu>
+  </menu>
+</extension>
+<extension point="xbmc.addon.metadata">
+  <summary lang="en_GB">TheMovieDb Helper - Now With Trakt Support!</summary>
+  <description lang="en_GB">TheMovieDb Helper provides details about movies, tvshows and actors from TMDb. Users can access a variety of lists from TMDb and Trakt.</description>
+  <disclaimer lang="en_GB">Provided by TheMovieDb and Trakt.</disclaimer>
+  <license>GPL-3.0</license>
```
enen92, line 68: this is not a valid spdx identifier, see https://spdx.org/licenses/
Should be GPL-3.0-only or GPL-3.0-or-later

plugin.video.themoviedb.helper/resources/lib/downloader.py
```diff
@@ -0,0 +1,138 @@
...
+    from urllib.parse import urlparse
+except ImportError:  # Python 2
+    from urlparse import urlparse
+
+
+class Downloader(object):
+    def __init__(self, download_url=None, extract_to=None):
+        self.addon = xbmcaddon.Addon()
+        self.download_url = download_url
+        self.extract_to = xbmc.translatePath(extract_to)
```
enen92, line 20: note: I'll probably move this to `xbmcvfs` at some point

plugin.video.themoviedb.helper/resources/lib/player.py
```diff
@@ -0,0 +1,393 @@
...
+
+        # Play/Search found item
+        if player and player[1]:
+            action = string_format_map(player[1], self.item)
+            if player[0] and action.endswith('.strm'):  # Action is play and is a strm so PlayMedia
+                utils.kodi_log(u'Player -- Found strm.\nAttempting PLAYMEDIA({})'.format(action), 2)
+                xbmc.executebuiltin(utils.try_decode_string(u'PlayMedia({0})'.format(action)))
+            elif player[0]:  # Action is play and not a strm so play with player
+                utils.kodi_log(u'Player -- Found file.\nAttempting to PLAY: {}'.format(action), 2)
+                xbmc.Player().play(action, ListItem(library='video', **self.details).set_listitem())
```
enen92, line 198: plugins should not use xbmcPlayer to request playback. What is this for?

enen92: I may have asked before

plugin.video.themoviedb.helper/resources/lib/script.py
```diff
@@ -0,0 +1,490 @@
...
+    def del_path(self):
+        if self.home.getProperty(self.prefixlock):
+            self.added_path = None
+            self.unlock_path()
+            return
+
+        self.prev_path()
+        self.call_window()
+
+    def play(self):
```
enen92, line 328: same here. When is this called? For which type of items? Calling videodb:// or plugin:// paths with this method is fine. Getting remote streams and request playback via this method is not okay

plugin.video.themoviedb.helper/resources/lib/service.py
```diff
@@ -0,0 +1,460 @@
...
+
+                ''' Skin images will be tried to be accessed directly. For all other ones
+                    the source will be copied to the addon_data folder to get access.
+                '''
+                if xbmc.skinHasImage(image):
+                    if not image.startswith('special://skin'):
+                        image = os.path.join('special://skin/media/', image)
+
+                    try:  # in case image is packed in textures.xbt
+                        img = Image.open(xbmc.translatePath(image))
```
enen92, line 114: not closing the `Image` handler?

plugin.video.themoviedb.helper/resources/lib/service.py
```diff
@@ -0,0 +1,460 @@
...
+
+                    except Exception:
+                        return ''
+
+                else:
+                    targetfile = os.path.join(targetpath, filename)
+                    if not xbmcvfs.exists(targetfile):
+                        xbmcvfs.copy(image, targetfile)
+
+                    img = Image.open(targetfile)
```
enen92, line 125: same

plugin.video.themoviedb.helper/resources/lib/utils.py
```diff
@@ -0,0 +1,385 @@
...
+    filename = str(unicodedata.normalize('NFD', filename).encode('ascii', 'ignore').decode("utf-8"))
+    filename = ''.join(c for c in filename if c in VALID_FILECHARS)
+    return filename
+
+
+def makepath(path):
+        if xbmcvfs.exists(path):
+            return xbmc.translatePath(path)
+        xbmcvfs.mkdirs(path)
+        return xbmc.translatePath(path)
```
enen92, line 44: note: this might be moved to xbmcvfs

plugin.video.themoviedb.helper/resources/lib/script.py
jurialmunkey: > Calling videodb:// or plugin:// paths with this method is fine.

That's exactly what it is for.  User can define player files for other addons e.g. Netflix  with defined rules for how to play the item.  When the user clicks an item to play in TMDbHelper, the plugin uses JSONRPC to walk the paths of the external addon to find a playable item according to matching rules. The plugin file path is then passed to the player and combined with the details grabbed from online.

The main reason this method is necessary is because some addons provide playable items without detailed information. Passing to Player() is the only way to combine a playable path with listitem details. Also there are some scenarios where passing the plugin path to  `PlayMedia()` causes a crash.

TMDbHelper is not providing any playable content - it is just providing a way to interface with other addons.

plugin.video.themoviedb.helper/resources/lib/player.py
jurialmunkey: TMDbHelper doesn't provide any playable content. It has user definable "player" files which provides rules about how to interact with other addons that do provide content. When a user clicks on an item, they are presented with a dialog to choose which "player" to use. JSONRPC is then used to walk a defined plugin path according to a set of rules. When a matching item is found, that plugin path is passed to Player() and combined with the detailed information that TMDbHelper has pulled from online APIs.

There's more detail in the wiki:
https://github.com/jurialmunkey/plugin.video.themoviedb.helper/wiki/Player-Function

plugin.video.themoviedb.helper/resources/lib/service.py
jurialmunkey: Whoops! 
Closing here now:
https://github.com/xbmc/repo-plugins/blob/4f1a0b77afb48e4400e61c82fff9d7691a3a0b85/plugin.video.themoviedb.helper/resources/lib/service.py#L74

plugin.video.themoviedb.helper/resources/lib/service.py
jurialmunkey: Same as above. Closing here now:
https://github.com/xbmc/repo-plugins/blob/4f1a0b77afb48e4400e61c82fff9d7691a3a0b85/plugin.video.themoviedb.helper/resources/lib/service.py#L74

plugin.video.themoviedb.helper/addon.xml
jurialmunkey: Done.
https://github.com/xbmc/repo-plugins/blob/4f1a0b77afb48e4400e61c82fff9d7691a3a0b85/plugin.video.themoviedb.helper/addon.xml#L68

plugin.video.themoviedb.helper/addon.xml
jurialmunkey: Done.

plugin.video.themoviedb.helper/resources/lib/downloader.py
jurialmunkey: Makes sense to be under xbmcvfs

jurialmunkey:
@enen92 - Okay. I think I've addressed all your comments now.

plugin.video.themoviedb.helper/addon.xml
basrieter: They could now use a single library in combination with arguments. See here for an example: https://github.com/retrospect-addon/plugin.video.retrospect/blob/kodi-matrix/addon.xml. 

Especially here it would simplify the code and reduce the number of 'passthrough` python files. All the `context-*.py` could be put in a single file. I would say that would be preferred.

jurialmunkey: I didn't even know this was possible - is this a Matrix addition? My only concern would be that I'm trying to keep things compatible across Leia/Matrix as much as possible (at least for the time being) to minimize how much work needs to be done on a separate branch. However, if it works in Leia too, then I'll definitely change to this approach because it is much better.

enen92: Yeah Matrix only, I'd say to keep it like it is until you ditch leia :)

plugin.video.themoviedb.helper/.gitignore
```diff
@@ -0,0 +1,63 @@
+# Created by .gitignore support plugin (hsz.mobi)
```
enen92, line 1: please remove from submission

plugin.video.themoviedb.helper/addon.xml
basrieter: After we included this in Matrix, I created at patch that I apply to the Leia version to get a Matrix version with these menu change.

jurialmunkey: I decided to bite the bullet and make the change. I doubt I'm going to add many if any more context menu entries - and if I do, the merge between branches should be relatively clean.

plugin.video.themoviedb.helper/.gitignore
jurialmunkey: Whoops! Thought I already did. Must've only deleted from the Leia version.

Done!

jurialmunkey:
@enen92 - Okay, should be all good to go now! :)

jurialmunkey:
@enen92 - Thanks! Can you also merge my Leia version in this PR: https://github.com/xbmc/repo-plugins/pull/2849

It's the same code - just without the context menu argument changes, python dependency and LOGNOTICE instead of LOGINFO.

PR #2861, opened 2020-04-27
[plugin.audio.sportschau] 2.0.0

sarbes:
### Description
Sports video for ardmediathek.de and sportschau.de
### Checklist:


plugin.video.sportschau/default.py
```diff
@@ -0,0 +1,34 @@
+# -*- coding: utf-8 -*-
+import libard
+import resources.lib.jsonparser as jsonparser
+
+class sportschau(libard.libard):
```
enen92, line 5: Using 4 spaces instead of tabs is recommended by pep8 guidelines, probably a good idea to run reindent on your files at some point. Can be done on the next update

PR #2922, opened 2020-05-17
[plugin.audio.kxmxpxtx.bandcamp] 0.0.3

Virusmater:
### Description
v0.0.3  (2020-05-17)
                - Added wishlist
### Checklist:


plugin.audio.kxmxpxtx.bandcamp/default.py
```diff
@@ -140,11 +146,19 @@ def main():
             query = xbmcgui.Dialog().input(addon.getLocalizedString(30103))
         if query:
             search(query)
+    elif mode[0] == 'url':
+        url = args.get("url", None)[0]
+        build_song_list(*bandcamp.get_album_by_url(url))
+        xbmc.executeJSONRPC(
```
basrieter, line 152: Could you explain why this is needed?

Virusmater: Yatse devs requested API in order to support share from Bandcamp Android app to Kodi:
https://community.yatse.tv/t/send-to-kodi-from-bandcamp/2389
Calls are needed in order to autoplay first song from the playlist. This is the best way I came up with.
Ideally I need:
1. Make first item selected (tried with .select(True), it kinda worked, but looks like focus and selected item are different things in Kodi)
2. Get list item based on selected item
3. Extract url from list item
4. Start streaming
 I tried different things in order to set first item in focus and play it without json-rpc, but nothing worked for me.

Virusmater:
Yatse devs requested API in order to support share from Bandcamp Android app to Kodi:
https://community.yatse.tv/t/send-to-kodi-from-bandcamp/2389
Calls are needed in order to autoplay first song from the playlist. This is the best way I came up with.
Ideally I need:

1. Make first item selected (tried with .select(True), it kinda worked, but looks like focus and selected item are different things in Kodi)
2. Get list item based on selected item
3. Extract url from list item
4. Start streaming the url

I tried different things in order to set first item in focus and play it without json-rpc, but nothing worked for me.

plugin.audio.kxmxpxtx.bandcamp/default.py
enen92: Have you tried creating a xbmc.Playlist object, add all songs to the playlist and set the resolved url to the first item on the playlist?

Virusmater: Hey, it worked. Removed json-rpc hack. Add the commit to the PR. Thanks!

Virusmater:
Removed the hack with json-rpc.
Now a playlist created instead.

enen92:
I see what you want to do now, create a directory and start the autoplay. Unfortunately it can't be done this way because by using xbmc.Player().play() on listitems which have a path other than `plugin://` will force kodi to create a new entry in the database each time you play an item.

Can you  create an endpoint in your plugin that receives the song (or whatever content) id and sets the resolved url instead? Something similar to:

```
@route("/play/{songid}")
def play(id):
   liz = ListItem("My song 1")
   liz.setPath(resolve_url_for_stream_with_id(id))
   xbmcplugin.setResolvedUrl(handle, liz, true)
```

When you create your playlist, you point each of the items to this path:

```
for item in mylist:
    liz = ListItem("My playlist item")
    liz.setPath("plugin://my.plugin.foo/play/mysongid")
    playlist.add(liz)
```

Then you're free to use xbmc.Player().play() with the playlist. Each item that reaches the player will only be stored once and resolved by your plugin.

Virusmater:
This is happening already. In the code:
```
def build_song_list(band, album, tracks, autoplay=False):
    track_list = list_items.get_track_items(band=band, album=album, tracks=tracks)
    xbmcplugin.addDirectoryItems(addon_handle, track_list, len(track_list))
    xbmcplugin.setContent(addon_handle, 'songs')
    xbmcplugin.endOfDirectory(addon_handle)
    if autoplay:
        playlist = xbmc.PlayList(xbmc.PLAYLIST_MUSIC)
        playlist.clear()
        for url, list_item, folder in track_list:
            playlist.add(url, list_item)
        xbmc.Player().play(item=playlist)
```
variable url holds following type of string:
```
plugin://plugin.audio.kxmxpxtx.bandcamp/?url=https%3A%2F%2Ft4.bcbits.com%2Fstream%2F43f61b9e09bf284114a5db64785aab07%2Fmp3-128%2F1865857563%3Fp%3D0%26ts%3D1590321030%26t%3Dbb78ba1b0e1a3c0aec7763b9d27b3bf2beea4d3b%26token%3D1590321030_986d2995a320092ee714da1651c728df9bc15901&mode=stream&title=Feedbacker+-+%D0%92+%D0%94%D1%80%D1%83%D0%B3%D0%BE%D0%B9+%D0%A0%D0%B0%D0%B7+%28U-Lock+cover%29
```
so I'm adding bunch of plugin:// items into the playlist, not http(s)://
later on it's getting actually played by:
```
def play_song(url):
    play_item = xbmcgui.ListItem(path=url)
    xbmcplugin.setResolvedUrl(addon_handle, True, listitem=play_item)

def main():
    args = parse_qs(sys.argv[2][1:])
    mode = args.get('mode', None)
.........
    elif mode[0] == 'stream':
        play_song(args['url'][0])
```

enen92:
Okay, maybe I misread your code, but that's the way to go. 

I don't like the player.play() usage but we can't avoid it at the moment without support on the python api. I am still a bit puzzled on why you need to present a directory and play it as a playlist (why not only play it?) but I guess you have your reason for that. Hence, I'll not block the PR on that.

Virusmater:
Reason of creation folder next to the playlist that it isn't really possible to open playlist from the player! At least with the default skin only from remote/mouse.
https://forum.kodi.tv/showthread.php?tid=57758
Really strange to have such problems, but seems that an audio isn't a really priority for Kodi community.

PR #2929, opened 2020-05-20
[plugin.video.nos] 4.1.0

artofcode:
### Description

- Fixed playback error in newer Kodi versions because plugin didn't specify if target was audio or video
- Removed xbmcswift2 dependency
- Sorting added to category and video entries
- Updated code to work with Python 3

### Checklist:


plugin.video.nos/resources/lib/requirements.txt
```diff
@@ -1,4 +1,3 @@
-xbmcswift2
 requests
```
basrieter, line 1: Please don't submit this file at all to the Kodi repository

artofcode: Cheers, removed!

PR #2947, opened 2020-06-06
[plugin.video.liveleak] 1.4.1

Botster:
### Description
FIX: Gracefully handle anomalous missing element exceptions
Add: new categories since I needed to do a PR anyway
### Checklist:

### DIFF
--- a/plugin.video.liveleak/addon.xml
+++ b/plugin.video.liveleak/addon.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8" standalone="yes"?>
 <addon id="plugin.video.liveleak"
     name="LiveLeak"
-    version="1.4.0"
+    version="1.4.1"
     provider-name="Botster">
     <requires>
         <import addon="xbmc.python" version="2.24.0"/>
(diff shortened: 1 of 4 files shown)


plugin.video.liveleak/addon.xml
```diff
@@ -19,6 +19,6 @@
         <license>GNU General Public License Version 3</license>
```
enen92, line 19: Would be nice to change this to GPL-3.0-or-later (spdx identifier) in some future update

plugin.video.liveleak/default.py
```diff
@@ -220,11 +223,15 @@ def buildListItem(url_medium_meta):
...
+    # Consider possibly that a thumbnail src was not fetched
+    if thumbnail:
+        liz.setArt( {'thumb': thumbnail} )
     liz.setProperty('IsPlayable', 'true')
-    cmd = "XBMC.RunPlugin({})"
-    cmd = cmd.format( buildUrl( {'mode': 'label_user', 'user': credit} ) )
-    liz.addContextMenuItems([('Label user: %s' % credit, cmd)])
+    # If user not Unknown, allow labeling of distinct LL users
+    if credit != 'Unknown':
+        cmd = "XBMC.RunPlugin({})"
```
enen92, line 232: Would be nice to drop xbmc prefix here. They were already removed in matrix.

plugin.video.liveleak/default.py
```diff
@@ -322,19 +331,44 @@ def index(url):
     # Get list of individual posts from indexing page
     listing = bs(page, 'html.parser')
     posts = []
-    for item in listing.find_all('div', class_='featured_items_outer'):
+    try:
+        featured_items_outer = listing.find_all('div', class_='featured_items_outer')
+    except:
+        notify('Cannot find any Thread items')
```
enen92, line 337: Is notify a GUI element? If so the message needs to be localized

PR #2948, opened 2020-06-07
[plugin.video.pcloud-video-streaming] 1.4.2

gdomenici:
### Description
Fixed issue #27: In certain locales (such as en-nz), the "Back to parent folder" list item would sort after folders starting with a number. Now it sorts correctly at the top of the list.
### Checklist:

### DIFF
--- a/plugin.video.pcloud-video-streaming/addon.py
+++ b/plugin.video.pcloud-video-streaming/addon.py
@@ -153,7 +153,8 @@ if mode[0] in ("folder", "myshares"):
 			oneFileOrFolderItem = oneItem["metadata"]
 		filename = oneFileOrFolderItem["name"]
 		if oneFileOrFolderItem["isfolder"]:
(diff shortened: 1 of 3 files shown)


plugin.video.pcloud-video-streaming/addon.py
```diff
@@ -153,7 +153,8 @@ def AuthenticateToPCloud():
 			oneFileOrFolderItem = oneItem["metadata"]
 		filename = oneFileOrFolderItem["name"]
 		if oneFileOrFolderItem["isfolder"]:
-			li = xbmcgui.ListItem(filename, iconImage='DefaultFolder.png')
+			li = xbmcgui.ListItem(filename)
+			li.setIconImage("DefaultFolder.png")
```
enen92, line 157: a quick note in case you're not aware: this function has been removed in matrix. No problem in using it for helix-leia though.

PR #2960, opened 2020-06-21
[plugin.audio.radiobrowser] 1.2.0

segler-alex:
### Description
This is an update of the plugin audio radiobrowser to version 1.2.0 with the following changes:
* use new api as documented at https://api.radio-browser.info
* stop play before start again to fix play only every second time bug
* use xbmc log instead of print

### Checklist:

### DIFF
--- a/plugin.audio.radiobrowser/main.py
+++ b/plugin.audio.radiobrowser/main.py
@@ -22,6 +22,37 @@ my_stations = {}
 profile = xbmc.translatePath(addon.getAddonInfo('profile')).decode("utf-8")
 mystations_path = profile+'/mystations.json'
 
+import socket
+import random
+
+def get_radiobrowser_base_urls():
+    """
+    Get all base urls of all currently available radiobrowser servers
+
+    Returns: 
+    list: a list of strings
+
+    """
+    hosts = []
+    # get all hosts from DNS
+    ips = socket.getaddrinfo('all.api.radio-browser.info',
+                             80, 0, 0, socket.IPPROTO_TCP)
+    for ip_tupple in ips:
+        ip = ip_tupple[4][0]
+
+        # do a reverse lookup on every one of the ips to have a nice name for it
+        host_addr = socket.gethostbyaddr(ip)
+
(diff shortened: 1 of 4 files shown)


plugin.audio.radiobrowser/main.py
```diff
@@ -196,37 +262,40 @@ def delFromMyStations(stationid):
...
 
 elif mode[0] == 'play':
-    stationid = args['stationid'][0]
-    data = downloadFile('http://www.radio-browser.info/webservice/v2/json/url/'+str(stationid),None)
+    stationuuid = args['stationuuid'][0]
+    data = downloadApiFile('/json/url/' + str(stationuuid),None)
     dataDecoded = json.loads(data)
     uri = dataDecoded['url']
+    xbmc.Player().stop()
     xbmc.Player().play(uri)
```
enen92, line 278: Plugins should not call Player.play() but use xbmcplugin.setResolvedUrl instead. Can you adjust the add-on?

segler-alex: yes i will immediatelly

segler-alex: it does not work :(
maybe the wrong place to ask, but do you know of any reason why the addon handle can get the value of "-1" ?
addon_handle = int(sys.argv[1])

segler-alex: isFolder=True does make the handle transfer work again.... that is why it works for all other pages but the deepest one where the real streams are...

segler-alex: i found the solution, i was missing
li.setProperty('IsPlayable', 'true')

enen92: Yeah sorry for the delay. You need to set the IsPlayable property to true to make it work. The reason behind this request is that, if you don't do it, everytime you play a radio channel a new entry is added to the database. Plugin setresolvedurl assigns a unique path to each of the items, hence, you're always playing the same file although the resolved stream url might difer

PR #2961, opened 2020-06-22
[service.upnext@leia] 1.1.1

im85288:
### Add-on details:

- **General**
  - Add-on name: Up Next
  - Add-on ID: service.upnext
  - Version number: 1.1.1

- **Code location**
  - URL: https://github.com/im85288/service.upnext
  
A service add-on that shows a Netflix-style notification for watching the next episode. After a few automatic iterations it asks the user if he is still there watching.

A lot of existing add-ons already integrate with this service out-of-the-box.

### v1.1.1 (2020-06-21)
- Avoid conflict with external players
- Restore "Ignore Playlist" option
- Fix a known Kodi bug related to displaying hours
- Improvements to endtime visualization
- New translations for Hindi and Romanian
- Translation updates to Hungarian and Spanish

### v1.1.0 (2020-01-17)
- Add notification_offset for Netflix add-on
- Fix various runtime exceptions
- Implement new settings
- Implement new developer mode
- Show current time and next endtime in notification
- New translations for Brazilian, Czech, Greek, Japanese, Korean
- New translations for Russian, Slovak, Spanish, Swedish
- Translation updates to Croatian, French, German, Hungarian, Italian, Polish
        

### Checklist:


service.upnext/codecov.yml
```diff
@@ -0,0 +1,11 @@
+coverage:
```
enen92, line 1: please remove from the submission. If you're using kodi-addon-submitter, add this to .gitattributes

service.upnext/resources/lib/player.py
```diff
@@ -0,0 +1,58 @@
...
+
+    def is_tracking(self):
+        return self.state.track
+
+    def disable_tracking(self):
+        self.state.track = False
+
+    def onPlayBackStarted(self):  # pylint: disable=invalid-name
+        """Will be called when kodi starts playing a file"""
+        sleep(5000)  # Delay for slower devices, should really use onAVStarted for Leia
```
enen92, line 34: can you use waitForAbort here? Sleep time is quite high and this may cause kodi to hang if requested for abort

service.upnext/resources/lib/script.py
```diff
@@ -0,0 +1,110 @@
...
+        elif action == self.ACTION_NAV_BACK:
+            self.close()
+
+
+def test_popup(window):
+    popup = TestPopup(window, addon_path(), 'default', '1080i')
+    popup.show()
+    step = 0
+    wait = 100
+    timeout = 10000
```
enen92, line 91: not sure if this is used but an implementation with waitForAbort and abortRequested() is advised

service.upnext/codecov.yml
im85288: Thanks, added to .gitattributes

service.upnext/resources/lib/script.py
im85288: Only in testing, but will have a look or maybe exclude

service.upnext/resources/lib/player.py
im85288: Checking, it's been that way for some time..as the saying goes if it's not broke don't fix it! But yeah will have a look

service.upnext/resources/lib/script.py
dagwieers: This is a countdown counter for testing pop-up window for theme developers.
I don't see how we would implement this using `waitForAbort` and `abortRequested()`, or wh that is relevant here.

dagwieers: We are using `xbmc.sleep()` underneath, is that a problem as well?

dagwieers: So I don't think this is a problem, this is broken up into smaller pieces, totalling up to 10 seconds (the time we want the example pop-up to appear on screen).

im85288:
Closing and reopening as not sure how to force push using the addon submitter

service.upnext/resources/lib/script.py
enen92: If only used in testing there's no issue with it. There's also no issue if the big timeout chunk is split into smaller wait times as long as the request for abortion is checked in-between. This is usually not requested just for the sake of it, but as a way to avoid plugins/scripts hanging kodi on exit

PR #2967, opened 2020-06-22
remove service.upnext moved to repo-scripts

anxdpanic:
Added in https://github.com/xbmc/repo-scripts/pull/1599


anxdpanic:
@enen92 
Not sure if anything else needs to be done to move this

enen92:
Crap. Let me look into it when I get home. Thanks for noticing

PR #3030, opened 2020-07-22
[plugin.video.vtm.go] 1.1.0

michaelarnauts:
### Description
* IPTV Manager support
* Bugfixes for authentication and subtitles
* New channel (VTM Nieuws)
* Cleanup logging (Use `DEBUG` instead of `INFO`)

The full list of changes can be found here: https://github.com/add-ons/plugin.video.vtm.go/blob/master/CHANGELOG.md#v110-2020-07-22

### Checklist:

Additional information :


plugin.video.vtm.go/resources/lib/kodilogging.py
```diff
@@ -0,0 +1,50 @@
...
+            logging.ERROR: xbmc.LOGERROR,
+            logging.WARNING: xbmc.LOGWARNING,
+            logging.INFO: xbmc.LOGNOTICE,
+            logging.DEBUG: xbmc.LOGDEBUG,
+            logging.NOTSET: xbmc.LOGNONE,
+        }
+
+        # Map DEBUG level to LOGNOTICE if debug logging setting has been activated
+        # This is for troubleshooting only
+        if ADDON.getSetting('debug_logging') == 'true':
```
basrieter, line 35: I think this one is rather expensive to call for each log line? Would it not be better to call it in the constructor and set it the log level there?

plugin.video.vtm.go/addon.xml
```diff
@@ -22,8 +22,9 @@
         <disclaimer lang="nl_NL">Deze add-on wordt niet ondersteund door DPG Media en wordt aangeboden 'as is', zonder enige garantie. De VTM GO naam, VTM GO logo en de kanaallogo's zijn eigendom van DPG Media en worden gebruikt in overeenstemming met de fair use policy.</disclaimer>
         <platform>all</platform>
         <license>GPL-3.0</license>
-        <news>v1.0.2
-- Fixes authentication errors</news>
+        <news>v1.1.0
```
basrieter, line 25: There are much more changes in the add-on than just these two. Why not give the most important ones?

plugin.video.vtm.go/resources/lib/kodilogging.py
```diff
@@ -0,0 +1,50 @@
...
+        formatter = logging.Formatter("[{}] [%(name)s] %(message)s".format(ADDON.getAddonInfo("id")))
+        self.setFormatter(formatter)
+
+    def emit(self, record):
+        """ Emit a log message """
+        levels = {
+            logging.CRITICAL: xbmc.LOGFATAL,
+            logging.ERROR: xbmc.LOGERROR,
+            logging.WARNING: xbmc.LOGWARNING,
+            logging.INFO: xbmc.LOGNOTICE,
```
enen92, line 28: Doesn't exist in matrix, please use loginfo

plugin.video.vtm.go/resources/lib/vtmgo/vtmgoepg.py
```diff
@@ -78,6 +82,8 @@ class VtmGoEpg:
     """ VTM GO EPG API """
     EPG_URL = 'https://vtm.be/tv-gids/api/v2/broadcasts/{date}'
 
+    EPG_NO_BROADCAST = 'Geen uitzending'
```
enen92, line 85: Needs to be localised

plugin.video.vtm.go/addon.xml
michaelarnauts: Most of the merge requests are for IPTV Manager support. One improves subtitles, and another fixes a potential issue when tokens would be invalidated (but that hasn't happened up till now). The rest are for improved testing (but these aren't in this PR).

So besides IPTV Manager support, and improved subtitles, there isn't much other stuff for the users.

enen92 (changes requested):

plugin.video.vtm.go/resources/lib/vtmgo/vtmgoepg.py
michaelarnauts: No, this is the exact string we get from the EPG api to indicate that there is no data. This isn't shown in the UI.

plugin.video.vtm.go/resources/lib/kodilogging.py
michaelarnauts: Does LOGINFO also exist in Leia? I haven't published this addon to Matrix yet.

plugin.video.vtm.go/resources/lib/kodilogging.py
michaelarnauts: I've benchmarked this on my laptop:

```py
        start = time.time()
        for x in range(1000):
            ADDON.getSetting('debug_logging')
        end = time.time()
        _LOGGER.error('time = %s', end-start)

```

This returns:
```
2020-07-22 23:02:03.015 T:140123968890624   ERROR: [plugin.video.vtm.go] [menu] time = 0.00075888633728

``` 

This seems perfectly fine. If I do this in the constructor, I need to have code to invalidate this when the settings change.

plugin.video.vtm.go/resources/lib/kodilogging.py
michaelarnauts: I can't change this yet for Kodi 18, since `LOGINFO` [doesn't seem to be logged](https://forum.kodi.tv/showthread.php?tid=353818&pid=2944096#pid2944096) to file in Leia. That's why I do the mapping from DEBUG to LOGINFO during troubleshooting.

I will do this for Kodi Matrix though.

enen92: yeah it looks like it was introduced only for matrix, please keep it in mind and disregard for now. I pointed this out as I've see some if/else clauses along the code already focusing on matrix.

plugin.video.vtm.go/resources/lib/modules/iptvmanager.py
```diff
@@ -0,0 +1,109 @@
...
+        self._vtm_go = VtmGo(self._kodi)
+        self._vtm_go_epg = VtmGoEpg(self._kodi)
+        self.port = port
+
+    def via_socket(func):  # pylint: disable=no-self-argument
+        """Send the output of the wrapped function to socket"""
+
+        def send(self):
+            """Decorator to send over a socket"""
+            sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
```
enen92, line 35: I'd say it's probably a better idea to place the connect call inside the try block and add an exception too in case this fails. But it is fine as is, it's not our responsibility to ask for these kind of changes.

enen92 (approved):

enen92:
@basrieter good to go?

plugin.video.vtm.go/resources/lib/kodilogging.py
michaelarnauts: The code should be almost compatible with Matrix, it's the plan to have one codebase for both Matrix and Leia (hence the if/else cases). I will modify it to use LOGINFO for Matrix in a future release. Thanks for the heads-up anyway!

plugin.video.vtm.go/resources/lib/kodilogging.py
basrieter: I had performance issues with my Raspberry Pi's due to this kind of code. So it's not wrong, just a point of attention.

PR #3045, opened 2020-08-06
[plugin.whereareyou] 0.4.0+matrix+1

pkscout:
### Description
- added ability to switch to external apps using Harmony Hub
- fix for looping dialog box if called from home window

### Checklist:

Additional information :


plugin.whereareyou/resources/lib/whereareyou.py
```diff
@@ -1,48 +1,179 @@
...
+                    activity, cmds = self._get_mapping_details( json_mappings, item )
+                    self.MESSAGE = '%s\n%s' % (self.MESSAGE, self.SETTINGS['ADDONLANGUAGE']( 32303 ))
+                    use_extended_dialog = True
+                    break
+        if use_extended_dialog:
+            if self.DIALOG.yesno( self.TITLE, self.MESSAGE ):
+                self._run_activity( activity, cmds )
+        else:
+            self.DIALOG.ok( self.TITLE, self.MESSAGE )
+        xbmc.Player().play( os.path.join( self.SETTINGS['ADDONPATH'], 'resources', 'blank.mp4' ) )
```
enen92, line 57: Would be better to have this point to a plugin:// path that could later be resolved

pkscout: Unfortunately if I change that line to use plugin://plugin.whereareyou/resources/blank.mp4 it crashes Kodi (although there is nothing obvious in the logs as to why). I'm open to suggestions on how to address that.  You might have noticed in the original version I used xbmcplugin.setResolvedUrl() to set a path to play, and that works if you're in library mode. But if you do it from the home screen, Kodi ends up in a loop constantly calling the addon over and over. It looks like the home screen call generates a playlist (the library mode doesn't), and once an item is in that playlist, it just keeps playing it.  I tried like 10 different things to deal with this, and the only one that mostly works is what I submitted. This still generates a "next item in playlist not found" message on the home screen, but it's better than the alternatives. I'm only playing the video at all so Kodi doesn't complain that it couldn't play the video.  I guess this is what happens when you try and find a work around for stub files not showing messages with no optical drive (which I know is fixed in Matrix, but this addon now has some extra options, so I think it's still potentially useful in Matrix). '-)

enen92: thats fine for now. Kodi definitely needs a lot of work on the playlist handling

enen92 (approved):

PR #3056, opened 2020-08-11
[plugin.image.dumpert] 1.0.4+matrix.1

skipmodea1:
### Description
Made a python 3 version based upon my python 2 version.
### Checklist:

Additional information :


plugin.image.dumpert/addon.xml
```diff
@@ -0,0 +1,40 @@
...
+  <extension point="xbmc.addon.metadata">
+	<summary lang="en_GB">Watch funny pictures from Dumpert.nl (dutch)</summary>
+	<description lang="en_GB">Watch funny pictures from Dumpert.nl (dutch)</description>
+	<disclaimer lang="en_GB">For bugs, requests or general questions visit the Dumpert.nl thread on the Kodi forum.</disclaimer>
+	<summary lang="nl_NL">Bekijk grappige plaatjes van Dumpert.nl (dutch)</summary>
+	<description lang="nl_NL">Bekijk grappige plaatjes van Dumpert.nl (dutch)</description>
+	<disclaimer lang="nl_NL">Bugs of andere feedback op deze plugin kan geplaatst worden in de Dumpert.nl thread op het Kodi forum.</disclaimer>
+    <language>nl</language>
+    <platform>all</platform>
+    <license>GPL-2.0-or-later</license>
```
basrieter: The license type does not match the on in license.txt  and the plugin of the `krypton` branch?

basrieter (changes requested):
Please have a look at the license of this version. It seems to be different? If you update it, please use the PDX code (https://spdx.org/licenses/).

plugin.image.dumpert/addon.xml
skipmodea1: Yeah that's correct, i didn't do pull requests for the licence change. It is changed in my git for my addons however, so next time there is a real change, the licence change will be in the PR.

basrieter: The I would suggest to stick with the original license in the addon.xml and make the complete change in the next PR. Now they are inconsistent, within Matrix, but also between Matrix and other Kodi versions.

skipmodea1: Do i need to change something?

micahg:
@basrieter not sure what the problem is (sorry). That page shows "GNU General Public License v2.0 or later" having a code of GPL-2.0-or-later, which is what I stuck in the addon.xml.

basrieter:
The issue is that the Kodi Matrix version of the Dumpert add-on has a different license than the Dumpert add-on from per-Matrix. They should be the same, as it is the same add-on, just targeted at a new Kodi version. The LICENSE.txt still shows GPL3, just like the pre-Matrix version. So the addon.xml should als have GPL3 targeted: either GPL-3.0-only or GPL-3.0-or-later. 

So the addon.xml needs to be updated to reflect the correct license.txt.

plugin.image.dumpert/addon.xml
basrieter: The issue is that the Kodi Matrix version of the Dumpert add-on has a different license than the Dumpert add-on from per-Matrix. They should be the same, as it is the same add-on, just targeted at a new Kodi version. The LICENSE.txt still shows GPL3, just like the pre-Matrix version. So the addon.xml should als have GPL3 targeted: either GPL-3.0-only or GPL-3.0-or-later.

So the addon.xml needs to be updated to reflect the correct license info.

skipmodea1: addon.xml is changed.

basrieter (approved):

basrieter:
Thanks, merged it.

PR #3193, opened 2020-11-01
[plugin.audio.pureradio] 2.0.0

bdommel:
Update for Kodi 19+
- python 3 support
- podcast section update


learningit:
Please pick and squash the 12 commits into a single commit before we review

learningit:
It will probably be easier for you if you create a new branch from matrix in the forked repo. Name it something like "matrix-plugin.audio.pureradio-2.0.0" so it has meaning.

PR #3195, opened 2020-11-01
[plugin.audio.pureradio] 2.0.0

bdommel:
update for Kodi 19 and higher
Python3 added
Podcast section updated


plugin.audio.pureradio/purelib.py
```diff
@@ -0,0 +1,150 @@
...
+
+def playStream():
+    # music item
+    pl = xbmc.PlayList(xbmc.PLAYLIST_MUSIC)
+    # clear playlist
+    pl.clear()
+    # add url item 
+    pl.add(URL)
+    # play the item
+    xbmc.Player().play(pl)
```
learningit, line 45: To use xbmc.Player().play() for a playlist in a plugin requires a bit more work. Direct urls should not be in the playlist. Playlist items need to be of the form plugin://......... with the url as a parameter so that the plugin gets called to play the url. The play routine should make use of  xbmcplugin.setResolvedUrl() to actually play the playable url.

xbmc.Player().play() should not be used to play a url in a plugin. It should be used in a script.

plugin.audio.pureradio/purelib.py
```diff
@@ -0,0 +1,150 @@
...
+    li = xbmcgui.ListItem(name)
+    # add properties
+    li.setInfo('music', {'Title': name})
+    li.setProperty('mimetype', 'audio/mpeg')
+    li.setProperty('IsPlayable', 'true')
+    li.setArt({'Icon': icon})
+    # create playlist item
+    pl = xbmc.PlayList(xbmc.PLAYLIST_MUSIC)
+    # clear playlist
+    pl.clear()
```
learningit, line 112: same issue here as described previously.

plugin.audio.pureradio/addon.xml
```diff
@@ -0,0 +1,26 @@
...
+		<provides>audio</provides>
+	</extension>	
+	<extension point="xbmc.addon.metadata">
+		<summary lang="en_GB">Pureradio.One</summary>
+		<description lang="en_GB">Pureradio.One is an internet radiostation which plays music from the 60's till today, commercial free and in great quality.</description>
+		<summary lang="nl_NL">Pureradio.One</summary>
+		<description lang="nl_NL">Pureradio.One is een internet radiostation met een brede mix van de beste hits, wereldwijd te volgen via een van onze streams.</description>
+		<platform>all</platform>
+		<license>GPL-2.0-only</license>
+		<assets>
```
learningit, line 16: it would be good to have <forum> and <source> entries here.

bdommel:
I will do a rewrite with the suggestions above.

PR #3200, opened 2020-11-03
[plugin.audio.pureradio] 3.0.0

bdommel:
Update for Kodi matrix
- code rewrite, based on current Kodi addon guidelines
- dynamic streaming and podcast urls
- update podcast section


learningit:
This looks a lot better, you're nearly there. The problem with this is the PR is raised against the master branch. You need to specify the branch (I think you want matrix) when creating the PR. You also need to remove the .pyc files from the PR. A quick look at everything else looks ok to me.

learningit:
Oh, also your PR has version 3.0.0 but your branch has version 3.0.1. Which is it? Both should match whatever is in addon.xml

bdommel:
I will do the commit to matrix, and will adjust the version number.

Thank you very much for the help!

PR #3201, opened 2020-11-03
[plugin.video.streamz@leia] 1.0.0

michaelarnauts:
### Add-on details:

- **General**
  - Add-on name: Streamz
  - Add-on ID: plugin.video.streamz
  - Version number: 1.0.0
  - Kodi/repository version: leia

- **Code location**
  - URL: https://github.com/add-ons/plugin.video.streamz
  
Streamz is a video-on-demand platform offering Flemish content from DPG Media, Telenet and VRT, but also including 'Home of HBO' content and blockbuster movies.

### Description of changes:

v1.0.0 (2020-11-03)
- Initial release

### Checklist:

- [x] My code follows the [add-on rules](http://kodi.wiki/view/Add-on_rules) and [piracy stance](http://kodi.wiki/view/Official:Forum_rules#Piracy_Policy) of this project. 
- [x] I have read the [CONTRIBUTING](https://github.com/xbmc/repo-scripts/blob/master/CONTRIBUTING.md) document
- [x] Each add-on submission should be a single commit with using the following style: [plugin.video.foo] v1.0.0

PR #3207, opened 2020-11-11
[plugin.video.vtm.go@leia] 1.2.0

michaelarnauts:
### Add-on details:

- **General**
  - Add-on name: VTM GO
  - Add-on ID: plugin.video.vtm.go
  - Version number: 1.2.0

- **Code location**
  - URL: https://github.com/add-ons/plugin.video.vtm.go
  
This add-on gives access to all live tv channels and all video-on-demand content available on the VTM GO platform.

### Description of changes:

v1.2.0 (2020-11-11)
- Implement Kodi Library support.
- Use new Movies and Series categories.
- Bugfixes and improved error handling.

### Checklist:

### DIFF
--- a/plugin.video.vtm.go/resources/lib/addon.py
+++ b/plugin.video.vtm.go/resources/lib/addon.py
@@ -6,9 +6,10 @@ from __future__ import absolute_import, division, unicode_literals
 import logging
 
 import routing
+from requests import HTTPError
 
-from resources.lib import kodiutils
-from resources.lib import kodilogging
+from resources.lib import kodilogging, kodiutils
+from resources.lib.vtmgo.exceptions import InvalidLoginException, LoginErrorException
 
 kodilogging.config()
 routing = routing.Plugin()  # pylint: disable=invalid-name
@@ -19,13 +20,39 @@ _LOGGER = logging.getLogger('plugin')
 @routing.route('/')
 def index():
     """ Show the profile selection, or go to the main menu. """
-    if not kodiutils.has_credentials() or (kodiutils.get_setting_bool('auto_login') and bool(kodiutils.get_setting('profile'))):
-        # If we have no credentials (browse only mode) or we have autologin and a profile, go directly to the main menu
-        show_main_menu()
+    while True:
+        if not kodiutils.has_credentials():
+            if not kodiutils.yesno_dialog(message=kodiutils.localize(30701)):  # You need to configure your credentials...
+                # We have no credentials
+                kodiutils.end_of_directory()
+                kodiutils.execute_builtin('ActivateWindow(Home)')
+                return
 
-    else:
-        # Ask the user for the profile to use
-        select_profile()
+            kodiutils.open_settings()
+        else:
+            break
+
+    try:
+        if kodiutils.get_setting_bool('auto_login') and kodiutils.get_setting('profile'):
+            # We have a profile
+            show_main_menu()
+
+        else:
+            # Ask the user for the profile to use
+            select_profile()
+
+    except InvalidLoginException:
+        kodiutils.ok_dialog(message=kodiutils.localize(30203))  # Your credentials are not valid!
+        kodiutils.open_settings()
+        kodiutils.end_of_directory()
+
+    except LoginErrorException as exc:
+        kodiutils.ok_dialog(message=kodiutils.localize(30702, code=exc.code))  # Unknown error while logging in: {code}
+        kodiutils.end_of_directory()
+
+    except HTTPError as exc:
+        kodiutils.ok_dialog(message=kodiutils.localize(30702, code='HTTP %d' % exc.response.status_code))  # Unknown error while logging in: {code}
+        kodiutils.end_of_directory()
 
 
 @routing.route('/menu')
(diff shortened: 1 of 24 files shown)


plugin.video.vtm.go/resources/lib/addon.py
```diff
@@ -19,13 +20,39 @@
 @routing.route('/')
 def index():
     """ Show the profile selection, or go to the main menu. """
-    if not kodiutils.has_credentials() or (kodiutils.get_setting_bool('auto_login') and bool(kodiutils.get_setting('profile'))):
-        # If we have no credentials (browse only mode) or we have autologin and a profile, go directly to the main menu
-        show_main_menu()
+    while True:
```
enen92, line 23: this has the chance of locking kodi completely isn't it? What if someone closes kodi while in the loop? I think you need to check for abortRequest here

plugin.video.vtm.go/resources/lib/modules/library.py
```diff
@@ -0,0 +1,195 @@
...
+    @staticmethod
+    def configure():
+        """ Configure the library integration. """
+        # There seems to be no way to add sources automatically.
+        # * https://forum.kodi.tv/showthread.php?tid=228840
+
+        # Open the sources view
+        kodiutils.execute_builtin('ActivateWindow(Videos,sources://video/)')
+
+    @staticmethod
```
enen92, line 165: I know we don't have any better option to support addon content in the library at the moment. I hope that you guys will remove all these code/hacks when (and if) montellese's library integration features are supported.

enen92 (commented):
I didn't spot any issues with the code. As this adds library integration I'd like to have a second opinion from somebody else in the team

plugin.video.vtm.go/resources/lib/modules/library.py
michaelarnauts: I would love to do that, it took me quite some time to figure out a way that works.

plugin.video.vtm.go/resources/lib/addon.py
michaelarnauts: We are asking the user for confirmation, and if he doesn't want to enter the credentials, we `return` out of the loop. The loop is just to ask the user to update his credentials again when they are incorrect.

I've tried to stop kodi, and it terminates just fine (without a timeout). I think Kodi is handling this since we are waiting for user input with the `yesno_dialog`.

michaelarnauts:
@enen92 the same code for Library is in the Addon for Streamz here: https://github.com/xbmc/repo-plugins/pull/3201

plugin.video.vtm.go/resources/lib/addon.py
learningit: I tend to agree with @enen92 on his comment here. I'm not sure how to test to confirm this though.

learningit:
A general note: not sure if it's by design or not but when I look at say kodiutils.py I see a global import xbmcgui but further down in the code there are Dialog imports from xbmcgui. It does no harm but is odd, considering the import of json is postponed several places for performance (I'm guessing). I've asked others about this because I do it myself in my library module and have been told it's bad coding practice. Nothing here that affects the review just a couple of informational points.

learningit:
The rest looks good to go

plugin.video.vtm.go/resources/lib/addon.py
michaelarnauts: I've installed the addon without credentials and opened it. It popped up the question to fill in the details. Pressing esc, back or "No" stops the loop and returns to the Homescreen. I've also tried to send a SIGTERM signal to Kodi while the dialog was shown, and it gracefully stopped Kodi.

I don't see an issue here.

enen92: Nevertheless and to avoid black magic can you loop over abortrequested instead?

michaelarnauts:
> A general note: not sure if it's by design or not but when I look at say kodiutils.py I see a global import xbmcgui but further down in the code there are Dialog imports from xbmcgui. It does no harm but is odd, considering the import of json is postponed several places for performance (I'm guessing). I've asked others about this because I do it myself in my library module and have been told it's bad coding practice. Nothing here that affects the review just a couple of informational points.

The kodiutils is the wrapper for all Kodi functions. You always need stuff from xmbcgui, so that's why it's imported globally. It's actually a PIP guideline somewhere to always import globally at the top.

We sometimes do a selective import for Kodi Leia/Matrix compatibility. When the import for Matrix fails, we do an import for Leia instead.

learningit:
> > A general note: not sure if it's by design or not but when I look at say kodiutils.py I see a global import xbmcgui but further down in the code there are Dialog imports from xbmcgui. It does no harm but is odd, considering the import of json is postponed several places for performance (I'm guessing). I've asked others about this because I do it myself in my library module and have been told it's bad coding practice. Nothing here that affects the review just a couple of informational points.
> 
> The kodiutils is the wrapper for all Kodi functions. You always need stuff from xmbcgui, so that's why it's imported globally. It's actually a PIP guideline somewhere to always import globally at the top.
> 
> We sometimes do a selective import for Kodi Leia/Matrix compatibility. When the import for Matrix fails, we do an import for Leia instead.

I meant that the local import is unnecessary because of the global. Guess I wasn't clear. No harm done as I said just unnecessary.  Or maybe I didn't notice something I should have.

dagwieers:
@learningit Yes, we often do local imports to avoid the initial import of libraries that are not being used in all code paths. It does speed up the add-on. If you use `reuselanguageinvoke` this only affects the first run of the add-on (after switching back to it). But it also impacts starting a live stream, a movie or a tv show from the EPG (or channel hopping).

Some Python sanity checks consider this a bad practice, but Kodi's add-on implementation is an exception to the rule. I would even say that the way Kodi uses add-ons is also a bad practice ;-)

See: https://github.com/add-ons/plugin.video.vrt.nu/wiki/Philosophy#importing-libraries-on-raspberry-pi

PS It also affects some platforms more than others. Here you can find our rationale:

learningit:
> add-on

For the second time, I commented on the local import after a global import. It's unnecessary but doesn't hurt anything. Unless I'm reading the code incorrectly I see a global import of xbmcgui and a local import of Dialog from xbmcgui later on. It's always possible I'm not seeing something because I am only looking at the diff but I don't think that makes a difference. I also postpone importing some larger modules in my helper routines until I know they need to be used but that is not what I'm commenting on. I've also said that it's harmless 3 times.

dagwieers:
I was only replying to this:

> It does no harm but is odd, considering the import of json is postponed several places for performance (I'm guessing).

learningit:
> I was only replying to this:
> 
> > It does no harm but is odd, considering the import of json is postponed several places for performance (I'm guessing).

I guess my English is poor. I was comparing the postponement of the import of json vs the global then local import. My point being: it looks like at one point the import of Dialog from xbmcgui was deferred and later the global import of xbmcgui was added making the local import redundant, slightly different but unnecessary if you think about it. Unless I'm missing a class definition or something else which is entirely possible. It's really unimportant and I was just commenting on the redundancy.

dagwieers:
Correct, the original kodiutils is only importing xbmcgui locally.
→ https://github.com/add-ons/plugin.video.vrt.nu/blob/master/resources/lib/kodiutils.py

michaelarnauts:
@learningit I'm sorry, I misunderstood your comment. I see now what you mean, I'll remove the additional imports in a future version and just use the globally imported one.

plugin.video.vtm.go/resources/lib/addon.py
michaelarnauts: @enen92 I understand what you mean, but there isn't anything magic happening here.

I've looked into this but noticed that listening to the `abortRequested()` in the while won't even work. When a shutdown is requested, the `while not monitor.abortRequested():` will not even be evaluated since the the `xbmcgui.Dialog().yesno()` will return False and the addon will end anyway.

Here is the same code with a bit more comments, I hope this makes it more clear.
```py
    while True:
        if not kodiutils.has_credentials():
            if not kodiutils.yesno_dialog(message=kodiutils.localize(30701)):  # You need to configure your credentials...
                # The user has pressed Esc, back doesn't want to fill in credentials, or Kodi is terminating
                # Stop the Add-on
                
                kodiutils.end_of_directory()
                kodiutils.execute_builtin('ActivateWindow(Home)')
                return
           
            # The user selected Yes and wants to enter his credentials
            kodiutils.open_settings()
        else:
            # Great, we have valid credentials and want to continue outside the loop
            break

    # We have valid credentials and can continue
    
    # more code...
```

enen92 (approved):

PR #3234, opened 2020-12-05
[plugin.video.tvvlaanderen@leia] 1.0.0

michaelarnauts:
### Add-on details:

- **General**
  - Add-on name: TV Vlaanderen
  - Add-on ID: plugin.video.tvvlaanderen
  - Version number: 1.0.0

- **Code location**
  
This add-on gives access to the live tv channels and the video-on-demand content available in your TV Vlaanderen subscription.

### Description of changes:

v1.0.0
- First release.

### Checklist:


plugin.video.tvvlaanderen/addon.xml
```diff
@@ -0,0 +1,37 @@
...
+        <import addon="script.module.requests" version="2.22.0"/>
+        <import addon="script.module.routing" version="0.2.0"/>
+        <import addon="script.module.pyjwt" version="1.6.4"/>
+        <import addon="inputstream.adaptive" version="2.4.3"/>
+    </requires>
+    <extension point="xbmc.python.pluginsource" library="addon_entry.py">
+        <provides>video</provides>
+    </extension>
+    <extension point="xbmc.addon.metadata">
+        <summary lang="nl_NL">Bekijk Live en on-demand TV via je TV Vlaanderen abonnement.</summary>
```
enen92, line 17: Is this nl only? If so please set the language tag so it can be hidden for foreign users

plugin.video.tvvlaanderen/resources/lib/solocoo/auth.py
```diff
@@ -0,0 +1,388 @@
...
+            raise Exception('Invalid tenant: %s' % tenant)
+
+        self._token_path = token_path
+
+        # Load existing account data
+        self._account = AccountStorage()
+        self._load_cache()
+
+        # Generate device serial if we have none
+        if not self._account.device_serial:
```
enen92, line 100: Can you explain why you need to expose the user uuid to an external service?

enen92 (commented):
See comments

plugin.video.tvvlaanderen/resources/lib/solocoo/auth.py
michaelarnauts: To authenticate, we need to present an unique ID for the device to the backend. You can only login to a few devices at the same time, so we need to remember the id that we have generated.

The id we generate is random, and not tied to any system id or something.

plugin.video.tvvlaanderen/addon.xml
michaelarnauts: Hmm, there are also French and English channels, but it's a Belgian service (Belgium has 3 official languages). Is there a way to set a tag based on country?

enen92: Nop, only language. It's okay to keep it as is. Thanks.

PR #3267, opened 2020-12-21
[plugin.video.newyankeeworkshop] 0.0.3

th0ma5w:
### Description
  - Updated the plugin for Matrix to use Python 3.x
### Checklist:


plugin.video.newyankeeworkshop/addon.xml
```diff
@@ -0,0 +1,33 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<addon id="plugin.video.newyankeeworkshop" name="The New Yankee Workshop" version="0.0.3" provider-name="th0ma5w">
+	<requires>
+		<import addon="xbmc.python" version="3.0.0"/>
+		<import addon="xbmc.addon" version="18.9.701" />
```
enen92: please drop the xbmc.addon dependency

plugin.video.newyankeeworkshop/router.py
```diff
@@ -0,0 +1,62 @@
...
+    xbmcplugin.endOfDirectory(addon_handle)
+elif mode[0] == 'season':
+    season_link = args['season'][0]
+    episode_list = get_episode_list(season_link)
+    list_items=[]
+    for episode in episode_list:
+        episode_display, episode_link = episode
+        url = build_url({'mode': 'episode', 'season': season_link, 'episode': episode_link, 'display': episode_display})
+        li  = xbmcgui.ListItem(episode_display)
+        li.setArt({'icon': os.path.join(ADDON_PATH,'icon.png')})
```
enen92: maybe pass the whole dict to the setArt call? You'd avoid calling setArt multiple times which has a performance penalty.

enen92 (changes requested):
See comments. Thanks for the update

th0ma5w:
@enen92  All requested changes have been made. Thank you so much for the feedback, and I do think it is actually noticeably faster.

enen92 (approved):

PR #3293, opened 2020-12-30
[plugin.video.newyankeeworkshop] 0.0.3

th0ma5w:
### Description
Updates to video extraction code for Krypton and Leia
### Checklist:

### DIFF
--- a/plugin.video.newyankeeworkshop/addon.xml
+++ b/plugin.video.newyankeeworkshop/addon.xml
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8" standalone="yes"?>
-<addon id="plugin.video.newyankeeworkshop" name="The New Yankee Workshop" version="0.0.2" provider-name="th0ma5w">
+<addon id="plugin.video.newyankeeworkshop" name="The New Yankee Workshop" version="0.0.3" provider-name="th0ma5w">
 	<requires>
 		<import addon="xbmc.python" version="2.25.0" />
 		<import addon="xbmc.addon" version="13.0.0" />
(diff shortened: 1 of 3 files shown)


th0ma5w:
I am not sure of the version resolution logic. I would like to submit a Python 2 plugin for Krypton and Leia, but the version check is saying I already have the Python 3 version for Matrix. If the plugins share the same namespace, how can I submit fixes for the older platform releases? Thank you, and I sincerely apologize for the hassle.

enen92:
Hey @th0ma5w 
The error makes sense. If you are running Leia and have an add-on with version 0.0.3 installed and you later migrate to matrix, Kodi will need to update the add-on at Kodi startup (migration phase) to receive the python3 compliant version. For this to work, the version in matrix has to be higher than the version in Leia. So, if you have version 0.0.4 in matrix and version 0.0.3 in Leia it will work. 0.0.3 in both won't work. So you need to update the matrix version first with a higher version and then submit to Leia (a lower version).
What we have been doing is exploiting pep python local version identifiers to keep the "same" version numbers. So 0.0.3 in Leia (python2) and 0.0.3+matrix.1 (python3). The version is the same (0.0.3) but the version scheme makes it higher than the other one (0.0.3+matrix.x > 0.0.3).
To make it easy for devs to submit both versions at the same time we have worked some CI scripts/GitHub actions that will send a PR on your behalf to both versions whenever you tag a release from your branch. See the action here: https://github.com/xbmc/action-kodi-addon-submitter. It includes a kodi-matrix variable that will automatically rewrite addon.xml to create the local version identifier and submit to Leia too (given that the only changes are in the version and xbmc.python version numbers). You can check an example of use of such workflow in https://github.com/enen92/plugin.video.rtpplay/blob/master/.github/workflows/addon-submitter.yml.
This specific add-on is in Krypton and also in matrix, PRs handled automatically by GitHub actions from the upstream repo whenever I tag a release (git tag 0.0.3. && git push --tags)

Unfortunately we still need to document the usage a little better but a lot of people are already using it. If you need help with that just ping.
You can still do this manually. Update the add-on version in matrix to 0.0.3+matrix.1, and when that is merged and the repo syncs (~2 hours after merge) this PR will turn green when Travis task is run again.

Cheers

You can see an example

PR #3299, opened 2021-01-01
[plugin.video.svtplay@matrix] 5.1.12+matrix.1

linqcan:
### Add-on details:

- **General**
  - Add-on name: SVT Play
  - Add-on ID: plugin.video.svtplay
  - Version number: 5.1.12+matrix.1

- **Code location**
  - URL: https://github.com/nilzen/xbmc-svtplay
  
With this addon you can stream content from SVT Play (svtplay.se).

### Description of changes:

- Added support for Kodi 19 (Matrix)
    

### Checklist:


plugin.video.svtplay/addon.xml
```diff
@@ -0,0 +1,29 @@
+<addon id="plugin.video.svtplay" name="SVT Play" provider-name="nilzen" version="5.1.12+matrix.1">
+  <requires>
+    <import addon="script.module.requests" version="2.22.0" />
```
linqcan, line 3: @enen92 Will this be a problem in Matrix? Do I need to update this myself somehow or should `kodi-addon-submitter` take care of this? 🤔

linqcan: `2.22.0+matrix.1` seems to be the version I need according to https://github.com/xbmc/repo-scripts/blob/matrix/script.module.requests/addon.xml ❓

enen92: It's fine as is. The version tag for a dependency specifies the minimum version required (2.22.0+matrix.1 is > than 2.22.0). When Kodi migrates the add-ons it starts by migrating the dependencies first, so you'll receive the matrix version of requests before receiving the version of stvplay. This was fixed and improved in https://github.com/xbmc/xbmc/pull/17703 if you're curious about the implementation.
Cheers

plugin.video.svtplay/resources/__init__.py
```diff
@@ -0,0 +1 @@
+# Dummy to declare python package
```
enen92, line 1: One note, when you stop supporting py2 and concentrate only on py3 you can safely remove all these empty init file placeholders. Python 3 doesn't require them

enen92 (approved):

linqcan:
@enen92 Thanks for all the help!

PR #3412, opened 2021-02-28
[plugin.video.rivedila7] 5.0.0

luivit:
### Description
version 5 for kodi matrix
### Checklist:


plugin.video.rivedila7/addon.xml
```diff
@@ -0,0 +1,36 @@
...
+    <provides>video</provides>
+  </extension>
+  <extension point="xbmc.addon.metadata">
+    <summary lang="en_GB">LA7-LA7d</summary>
+    <description lang="en_GB">Live Streaming, Replay the last week, LA7 Prime, On-Demand programs, News and Weather, Teche LA7-The Protagonists</description>
+    <summary lang="it_IT">LA7-LA7d</summary>
+    <description lang="it_IT">Diretta Live, Rivedi ultima settimana, LA7 Prime, Programmi On-Demand, Telegiornali e Meteo, Teche LA7-I Protagonisti</description>
+    <language>it</language>
+    <platform>all</platform>
+    <license>GNU General Public License v3.0</license>
```
enen92: GPL-3.0-only or GPL-3.0-or-later (spdx identifier)

enen92 (commented):
Minor, good to go after that.

plugin.video.rivedila7/addon.xml
Testato: may you made an example ? how i should change addon.xml for GPL-3.0-or-later ?

enen92: Just use "GPL-3.0-only" (or "GPL-3.0-or-later" if you prefer) instead of the full name of the license. No need for anything else.
See spdx.org

enen92: https://kodi.wiki/view/Addon.xml#.3Clicense.3E

enen92 (approved):

PR #3432, opened 2021-03-03
[plugin.video.lbry] 0.1.3

accumulator:
### Description
Update of plugin.video.lbry from Leia to Matrix

### Checklist:


plugin.video.lbry/resources/language/strings.pot
```diff
@@ -0,0 +1,144 @@
+msgid ""
```
enen92: this file is incorrect, remove it

plugin.video.lbry/resources/lib/local.py
```diff
@@ -0,0 +1,65 @@
...
+import xbmc
+import xbmcaddon
+import xbmcvfs
+import xbmcgui
+
+ADDON = xbmcaddon.Addon()
+tr = ADDON.getLocalizedString
+
+def get_profile_path(rpath):
+    return xbmc.translatePath(ADDON.getAddonInfo('profile') + rpath)
```
enen92:
```suggestion
    return xbmcvfs.translatePath(ADDON.getAddonInfo('profile') + rpath)
```

plugin.video.lbry/resources/lib/exception.py
```diff
@@ -0,0 +1,5 @@
+# -*- coding: utf-8 -*-
```
enen92, line 1: empty/dummy `__init__.py` files should no longer be needed in python3

plugin.video.lbry/resources/lib/local.py
```diff
@@ -0,0 +1,65 @@
...
+    for line in lines.split('\n'):
+        items = line.split('#')
+        if len(items) < 2:
+            continue
+        channels.append((items[0],items[1]))
+    return channels
+
+def save_channel_subs(channels):
+    try:
+        f = xbmcvfs.File(get_profile_path('channel_subs'), 'w')
```
enen92: jfyi `xbmcvfs.File` does support context lib, so you should be able to use:

```python
with xbmcvfs.File(get_profile_path('channel_subs'), 'w') as f:
    f.something()
``` 

Not mandatory to change though.

enen92 (changes requested):
See comments

accumulator:
@enen92 issues resolved

enen92 (approved):

PR #3456, opened 2021-03-13
[plugin.video.drnu] 5.0.15

TermeHansen:
### Description
update
- Add buggalo error submit to kodi.drnu gmail
- fix bug for write subtitle files on libreelec, kodi 18.9

### Checklist:

### DIFF
--- a/plugin.video.drnu/resources/lib/addon.py
+++ b/plugin.video.drnu/resources/lib/addon.py
@@ -61,6 +61,15 @@ class DrDkTvAddon(object):
         self.cache_path = xbmc.translatePath(xbmcaddon.Addon().getAddonInfo('profile'))
         if not os.path.exists(self.cache_path):
             os.makedirs(self.cache_path)
+        buggalo.EMAIL_CONFIG = {
+                 "recipient":"drnu.kodi@gmail.com",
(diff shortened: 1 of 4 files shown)


plugin.video.drnu/resources/lib/addon.py
```diff
@@ -61,6 +61,15 @@ def __init__(self, plugin_url, plugin_handle):
         self.cache_path = xbmc.translatePath(xbmcaddon.Addon().getAddonInfo('profile'))
         if not os.path.exists(self.cache_path):
             os.makedirs(self.cache_path)
+        buggalo.EMAIL_CONFIG = {
```
enen92, line 64: I checked with the team, can you please make this opt-in instead of enabled by default? Emailing the author without the user consent is not a great idea. Or is the user prompted to do it internally in buggalo?

TermeHansen: Yes they are prompted in buggalo and the mail is a mail made for the addon so hopefully future developers can take it over and also look back in old bug reports if needed

https://kodi.wiki/view/Add-on:Buggalo_Exception_Collector

enen92 (approved):
That's fine then, if users are prompted. Last time I used buggalo was really a long time ago :)

PR #3470, opened 2021-03-19
[repo-plugins][matrix] Add addon-checker via github actions

enen92:
This PR provides a common action to be added to each branch to run Kodi addon checker with github actions as an alternative to Travis.
It was written to mimic what is currently implemented in Travis and in a way it can be easily copied to each branch without requiring any adjustment or modification.

I'd like to have this approved first and then port to every other branch in all repos (repo-plugins, repo-scripts, repo-scrapers and repo-resources). 

The only thing left is something similar to travis buddy (which is broken atm anyway). I've tried adding it to the action but github does not allow it for security reasons on forked repos (see: https://github.com/thollander/actions-comment-pull-request/issues/10). I think the way to go is to use a standalone app we can host and configure the repo with webhooks.

Note we might want to add https://github.com/xbmc/repository-generator/pull/17 to the repository-generator at some point. It doesn't crash at the moment but it does log an exception, which can be a source of confusion in the future.

pinging @yol or @wsnipex 

At the moment the repo is blocked for new additions as we are out of travis credits. I'd like to have this in as soon as possible.

### DIFF
--- a/.github/workflows/addon-checker.yml
+++ b/.github/workflows/addon-checker.yml
@@ -1,0 +1,30 @@
+name: Kodi Addon-Checker
+
+on: [pull_request]
+
+jobs:
+  kodi-addon-checker:
+    runs-on: ubuntu-latest
+    name: Kodi Addon-Checker
+    steps:
+
+    - name: Checkout
+      uses: actions/checkout@v1
+
+    - name: Set up Python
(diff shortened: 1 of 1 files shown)


wsnipex (approved):
looks good to me

PR #3472, opened 2021-03-19
[plugin.video.rivedila7@matrix] 5.1.0

luivit:
### Description
- Fix On-Demand homepages scraping
- Removed minimal versions on import addon
### Checklist:

### DIFF
--- a/plugin.video.rivedila7/addon.xml
+++ b/plugin.video.rivedila7/addon.xml
@@ -1,11 +1,11 @@
 <?xml version="1.0" encoding="UTF-8" standalone="yes"?>
-<addon id="plugin.video.rivedila7" name="LA7-LA7d" version="5.0.0" provider-name="Luivit, Testato">
+<addon id="plugin.video.rivedila7" name="LA7-LA7d" version="5.1.0" provider-name="Luivit, Testato">
   <requires>
     <import addon="xbmc.python" version="3.0.0"/>
-    <import addon="script.module.requests" version="2.22.0"/>
-    <import addon="script.module.inputstreamhelper" version="0.5.0"/>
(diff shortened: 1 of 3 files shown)


plugin.video.rivedila7/addon.xml
```diff
@@ -1,11 +1,11 @@
 <?xml version="1.0" encoding="UTF-8" standalone="yes"?>
-<addon id="plugin.video.rivedila7" name="LA7-LA7d" version="5.0.0" provider-name="Luivit, Testato">
+<addon id="plugin.video.rivedila7" name="LA7-LA7d" version="5.1.0" provider-name="Luivit, Testato">
   <requires>
     <import addon="xbmc.python" version="3.0.0"/>
-    <import addon="script.module.requests" version="2.22.0"/>
-    <import addon="script.module.inputstreamhelper" version="0.5.0"/>
-    <import addon="script.module.beautifulsoup4" version="4.5.3"/>
-    <import addon="script.module.html5lib" version="0.999.0"/>
+    <import addon="script.module.requests" version="0.0.0"/>
```
enen92, line 5: why these changes?

enen92 (approved):

plugin.video.rivedila7/addon.xml
Testato: because i discover that this version declaration is only the Minimal version. Kodi will install always the last version.
So if on our plugin we have no minimal version required put 0.0.0 is a good choise.
If in the future we discover a bug on old version of some dependencies we increase this value.

I changed it also for a bug that i found on Windows, by declaring the correct last inputstreamhelper version here, windows not install It, and print an error on the version.

PR #3576, opened 2021-05-02
[plugin.audio.podcasts] 2.0.0

Heckie75:
### Description
Addon in order to subscribe to RSS podcasts. There is no service. The URLs of podcasts are provided within settings. It is also possible to use OPML files which can be exported by podcast receivers like AntennaPod.  

### Checklist:

Additional information :


plugin.audio.podcasts/addon.py
```diff
@@ -0,0 +1,340 @@
...
+
+        if settings.getSetting("label%s" % item_id) != "":
+            label = settings.getSetting("label%s" % item_id)
+
+        if "icon" in entry and entry["icon"].startswith("http"):
+            icon_file = entry["icon"]
+
+        elif "icon" in entry:
+            icon_file = os.path.join(addon_dir,
+                                     "resources",
```
basrieter: It seems there is not "assets" folder in "resources"?

basrieter (changes requested):
Please have a look at my comments and the results from the add-on checker: https://github.com/xbmc/repo-plugins/pull/3576/checks?check_run_id=2485977743

basrieter:
BTW: To update your PR, just add another commit to it (or ammend the current commit) and push the results. The PR will then be updated.

plugin.audio.podcasts/addon.py
Heckie75: Changed this so that there is now a default icon in case that no specific icon is given.

plugin.audio.podcasts/addon.py
```diff
@@ -0,0 +1,327 @@
...
+            items, image = self._load_rss_feed(url)
+            index = int(params["index"][0])
+
+            item = items[index]
+            if "enclosure" in item:
+                stream_url = item["enclosure"]["@url"]
+            else:
+                stream_url = item["guid"]
+
+            xbmc.executebuiltin('PlayMedia(%s)' % stream_url)
```
basrieter: I checked and a plugin should use xbmcplugin.setResolvedUrl() in combination with xbmcgui.ListItem.setProperty('IsPlayable', 'true'). That way Kodi will also remember what items were played in the past. But I guess this works. Same in other places.

plugin.audio.podcasts/addon.py
```diff
@@ -0,0 +1,327 @@
...
+        for entry in entries:
+
+            self._add_list_item(entry, path)
+
+        xbmcplugin.endOfDirectory(self._addon_handle)
+
+    def play(self, parent, path, params):
+
+        url = params["url"][0]
+        xbmc.executebuiltin('PlayMedia(%s)' % url)
```
basrieter: I checked and a plugin should use xbmcplugin.setResolvedUrl() in combination with xbmcgui.ListItem.setProperty('IsPlayable', 'true'). That way Kodi will also remember what items were played in the past. But I guess this works. Same in other places.

plugin.audio.podcasts/addon.py
basrieter: I still think this will not work. As there are no extra files in the "assets" folder. I guess this should be icon.png?

basrieter (changes requested):
Have a look at my comments and the `setResolvedUrl()` change. Als double check: https://github.com/xbmc/repo-plugins/pull/3576/checks?check_run_id=2487543619

plugin.audio.podcasts/addon.py
basrieter: Could you change this to use `xbmcplugin.setResolvedUrl()`

basrieter (changes requested):

plugin.audio.podcasts/addon.py
Heckie75: @basrieter I'm going to investigate what setResolvedUrl does and if it fits into my approach.

So far I have read this article: [What is the goal of the method "setResolvedUrl" ?](https://forum.kodi.tv/showthread.php?tid=173986)

Actually the URLs inside the plugin are like pointers to items in RSS feed, e.g. ˋˋˋpod-1/1ˋˋˋ which means first RSS feed and first entry in this feed. The advantage is that in this example the URL always points to the most recent podcast and therefore not to a speicific one. This is helpful when bookmarking it or using it in combination with my [timer addon](https://github.com/Heckie75/kodi-addon-heckies-timer) which I maybe want to post later.

As mentioned I will come back to you after further investigation and after I know in detail how setResolvedUrl works.

Heckie75:
@basrieter 
After hard investigation in terms of your remark to use setResolvedUrl I came to the conclusion that it is not appropriate for this addon. The reason is that the rss feed is dynamic. It changes with every new item that is posted in the podcast. Therefore it is required to query the rss feed and get the stream URLs each time you browse the directory. 

However, this gave my the chance of many refactorings. 

Pls. take a look a the latest commit.

basrieter:
> @basrieter
> After hard investigation in terms of your remark to use setResolvedUrl I came to the conclusion that it is not appropriate for this addon. The reason is that the rss feed is dynamic. It changes with every new item that is posted in the podcast. Therefore it is required to query the rss feed and get the stream URLs each time you browse the directory.
> 
> However, this gave my the chance of many refactorings.
> 
> Pls. take a look a the latest commit.

That should not be a problem at all. The setResolved() only tells Kodi what to what the item resolves so Kodi will play it. The benefit is that Kodi actually knows the media playback came from your stream and you could even add more meta data to it. This is the preferred way.

Heckie75:
@basrieter 
So may be this is what you ask for. Pls. double check latest commit.

plugin.audio.podcasts/addon.py
```diff
@@ -0,0 +1,359 @@
...
+            }
+        ]
+
+    def _play_latest(self, url):
+
+        try:
+            title, description, image, items = self._load_rss(url)
+            item = items[0]
+            li = self._create_playable_list_item(item)
+            xbmcplugin.setResolvedUrl(self.addon_handle, True, li)
```
basrieter, line 129: if you add this:

```
li.setProperty("IsPlayable", "true")
li.setProperty("path", item["stream_url"])
```

and remove 

``` 
xbmc.Player().play(item["stream_url"], li)
``` 

It should be as it is and Kodi should start playing the stream.

basrieter (changes requested):

plugin.audio.podcasts/addon.py
Heckie75: @basrieter 
I added `li.setProperty("path", item["stream_url"])` in `_create_playable_list_item` although I have already initialized the ListItem with `path`, see `li = xbmcgui.ListItem(label=item["name"], label2=item["name2"], path=item["stream_url"])`. But it seems to make a difference. 

`li.setProperty("IsPlayable", "true")` was already there, see  `_create_playable_list_item`

I removed `xbmc.Player().play(item["stream_url"], li)`

Last but not least there is a fix that was required in order to be able to navigate in Kodi's Android App (Core).

Heckie75: @basrieter 
Actually I see now that `xbmc.Player().play(item["stream_url"], li)` is still required since the stream doesn't start if this is missing. Any idea?

enen92: I guess it's something wrong with your routing logic. the playable listitem (the one with IsPlayable = True) should point to a "plugin://yourplugin/something" path and not to the final stream url. When that path is called, you then set the resolved url to the final path. The reason we suggest this approach is because it has implications on the database. This way your "path" is always the same (the plugin one) and you basically avoid adding new entries to the database everytime the stream changes (since it is dynamic)

basrieter: Instead of the `setProperty()` use `item.setPath()`. That should do the trick.

Heckie75: @basrieter 

It seems that `item.setPath` works after also changing that the addon provides audio and video and also set the media type.

I hope that the idea of bookmarking the latest stream in every podcast works. This is something that I will probably know by tomorrow.

basrieter: Shall I waith then for you to test?

Heckie75:
I will give you feedback this evening

plugin.audio.podcasts/addon.py
Heckie75: I have tested it and it works as exprected so that I can set a bookmark to the listitem that referes to the latest item in rss feed.

I found another bug that I've fixed that was caused by rss feeds with a single item.

basrieter:
> I will give you feedback this evening

It looks good to me. If it's done, I can merge it.

Heckie75:
From my point of view everything ist fine. If you don't ask for more changes, there won't be any (for this version) from my site. 

Many thanks for your effort!

basrieter (approved):

basrieter:
> From my point of view everything ist fine. If you don't ask for more changes, there won't be any (for this version) from my site.
> 
> Many thanks for your effort!

No worries. Thanks for your patience!

PR #3584, opened 2021-05-04
[plugin.video.rivedila7] 6.0.1

luivit:
### Description
plugin's code rework
### Checklist:

### DIFF
--- a/plugin.video.rivedila7/resources/lib/plugin.py
+++ b/plugin.video.rivedila7/resources/lib/plugin.py
@@ -4,948 +4,863 @@ import sys
 import xbmc
 import xbmcgui
 import xbmcplugin
-import xbmcaddon
 import urllib.request
 import urllib.parse
 from urllib.request import Request, urlopen
-import html5lib
 import time
 import requests
 import inputstreamhelper
 from bs4 import BeautifulSoup
 
-
-
-addon = xbmcaddon.Addon()
-language = addon.getLocalizedString
-handle = int(sys.argv[1])
-url_base = "https://www.la7.it"
-url_base_la7d = "https://www.la7.it/la7d"
-url_live_la7 = "https://www.la7.it/dirette-tv"
-url_live_la7d = "https://www.la7.it/live-la7d"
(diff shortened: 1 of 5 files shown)


plugin.video.rivedila7/.gitignore
```diff
@@ -0,0 +1,123 @@
+# Byte-compiled / optimized / DLL files
```
enen92: drop from the submission

plugin.video.rivedila7/resources/lib/plugin.py
```diff
@@ -4,948 +4,865 @@
...
     if link_video == '':
-        xbmc.log('NO VIDEO LINK',xbmc.LOGINFO)
-        if xbmcgui.Dialog().ok(addon.getAddonInfo('name'), language(32005)):
-            exit()
-    else:
-        listitem.setProperty('inputstream','inputstream.adaptive')
-        listitem.setProperty('inputstream.adaptive.manifest_type','hls')
-        listitem.setPath(link_video)
-        xbmcplugin.setResolvedUrl(handle, True, listitem)
+        xbmc.log('NO VIDEO LINK', xbmc.LOGINFO)
```
enen92: can you add some sort of plugin identifier to this log line? Anyone checking it on the log won't notice where it's coming from

enen92 (changes requested):
see comments, the .gitignore has to go

plugin.video.rivedila7/.gitignore
Testato: you mean that we need remove completely the.gitignore file ? or only this first line ?

enen92: All of it

Testato: done

plugin.video.rivedila7/resources/lib/plugin.py
Testato: done

enen92 (approved):

PR #3595, opened 2021-05-12
[plugin.audio.podcasts] 2.0.1

Heckie75:
### Description

Unfortunately the previous code uses a method `setDateTime` of listitem which is available in Kodi v20 but not in v19.x. This has been fixed to that addon also works with v19.x. 

### Checklist:

### DIFF
--- a/plugin.audio.podcasts/addon.py
+++ b/plugin.audio.podcasts/addon.py
@@ -139,9 +139,11 @@ class Mediathek:
         if "name2" in item:
             li.setProperty("label2", item["name2"])
 
-        if "stream_url" in item and "type" in item:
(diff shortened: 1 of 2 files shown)


Heckie75:
@basrieter 

Sorry, after I have installed the addon from the repository to a different PC and have seen that I have used a method that is not available in Kodi v19 yet. 

Therefore I changed the code so that it is compatible with both versions. However, sorting the directory by date won't be available in v19. Do you have any suggestions how to sort items by date in Kodi v19

It will be available in v20 automatically.

I also found that bookmarking of the  latest item works but not directly playing.

basrieter (approved):

PR #3618, opened 2021-05-24
[plugin.audio.bbcpodcasts] 1.0.0

Heckie75:
### Description
Podcasts by BBC in KODI. All contents are streamed from https://www.bbc.co.uk/podcasts

### Checklist:


plugin.audio.bbcpodcasts/resources/settings.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
sarbes, line 1: That's the old format for settings. See https://kodi.wiki/view/Add-on_settings_conversion.

sarbes: After confirming with my team members, this is not required for now. But please keep it in mind as it might be in the future.

Heckie75: Ok thanks, I am going to open an issue in my projects (also other addons) in order to change it later.

plugin.audio.bbcpodcasts/resources/lib/rssaddon/abstract_rss_addon.py
```diff
@@ -0,0 +1,277 @@
+from datetime import datetime
```
basrieter, line 1: Would it make sense to make a more generic module that you can then use in all your podcast/rss add-ons?

Heckie75: the way I've done it here works for me. I use a git submodule, see https://github.com/Heckie75/kodi-submodule-rssaddon, which I use in 4 addons, i.e.
- https://github.com/Heckie75/kodi-addon-podcast
- https://github.com/Heckie75/kodi-addon-deutschlandfunk
- https://github.com/Heckie75/kodi-addon-bbc-podcasts
- https://github.com/Heckie75/kodi-addon-ard-aktuell

(some of them are not updated in the repo yet)

I guess that more generic modules must be good enough so that other developers can use them with ease. I'm not sure if this is the case at this moment.

At least there would have to be a kind of manual.

What do you think?

basrieter: Well, if 4 add-ons use it, I would say that a generic module should be worth it. But that is up to you, perhaps for the future? For now this looks good.

basrieter (approved):

PR #3621, opened 2021-05-26
[plugin.video.ctuplinkrss] 1.2

oscaropenness:
### Description
update python version and dependencies to make add-on compatible with Kodi 19 -  Matrix
### Checklist:


plugin.video.ctuplinkrss/resources/lib/ctuplinkrss.py
```diff
@@ -0,0 +1,62 @@
...
+    
+    #Parse ctuplink-Feed
+    d = feedparser.parse('https://blog.ct.de/ctuplink/ctuplinkvideohd.rss')
+    
+    #Liste mit allen Folgen aufbauen
+    listing = []
+    
+    for item in d['entries']:
+        title = item['title']
+        url = item.enclosures[0].href
```
sarbes, line 34: I'm not sure if this might result in HEAD requests for every item added here. I'm fairly sure that older versions of Kodi do this, but I'm not certain about the current status.

oscaropenness: I am not sure, what you mean by this.
Are you saying the *feedparser* library is making HEAD requests when I call `item.enclosures[0].href`?
Would it be better to get the url by accessing `item.links[0].href`?

Or is Kodi making the request when I list them using `xbmcplugin.addDirectoryItems`?

Could you maybe point me to the documentation that would help me debug the unnecessary HEAD requests?

Thank you!

plugin.video.ctuplinkrss/resources/lib/ctuplinkrss.py
```diff
@@ -0,0 +1,62 @@
+#!/usr/bin/python
```
basrieter, line 1: Not needed.

basrieter: Also include the SPDX in each Python file.

plugin.video.ctuplinkrss/resources/lib/ctuplinkrss.py
ronie: should be good to go...
i've tested the addon and didn't spot any excessive HEAD requests.

ronie (approved):
looks good to me, thx!

PR #3642, opened 2021-06-16
[plugin.video.livestream@helix] 2021.6.16

eracknaphobia:
### Add-on details:

- **General**
  - Add-on name: Livestream
  - Add-on ID: plugin.video.livestream
  - Version number: 2021.6.16
  - Kodi/repository version: helix

- **Code location**
  - URL: https://github.com/eracknaphobia/plugin.video.livestream
  
Livestream, formerly known as Mogulus, is a live streaming video platform that allows users to view and broadcast video content using a camera and a computer through the internet.

### Description of changes:

        - Matrix compatibility updates
        - Code clean up
    

### Checklist:

### DIFF
--- a/plugin.video.livestream/README.md
+++ b/plugin.video.livestream/README.md
@@ -1,5 +1,10 @@
+[![GitHub release](https://img.shields.io/github/release/eracknaphobia/plugin.video.livestream.svg)](https://github.com/eracknaphobia/plugin.video.livestream/releases)
+![License](https://img.shields.io/badge/license-GPL%20(%3E%3D%202)-orange)
+[![Contributors](https://img.shields.io/github/contributors/eracknaphobia/plugin.video.livestream.svg)](https://github.com/eracknaphobia/plugin.video.livestream/graphs/contributors)
+
 plugin.video.livestream
 
 ======================
+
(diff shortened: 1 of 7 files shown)


eracknaphobia:
How are we supposed to handle this using kodi-addon-submitter to submit py2 & py3 versions when the older versions don't support strings.po language files?

`ERROR: Using the new language directory structure in /home/runner/work/repo-plugins/repo-plugins/plugin.video.livestream/resources/language/resource.language.en_gb for a Kodi version that does notsupport it. Please use the old language file struture or move the addon toan upper branch/kodi version.`

enen92:
Honestly the patch in addon-submitter to allow py2/py3 is more a workaround to allow an easy port to matrix (while still storing the addon in the leia branch) than anything that should be considered long term. I think the way to go is to drop support for those outdated platforms (at least anything < Leia). Hopefully when Nexus is out we can enforce submissions only for python3 (matrix and above)

enen92:
A possible approach is to use named branches like I do in https://github.com/XBMC-Addons/plugin.audio.radio_de

eracknaphobia:
Good point. Support for the older branches could probably be dropped at this point. As always thanks for the info @enen92!

PR #3702, opened 2021-08-15
[plugin.video.srf_ch_replay] 2.0.1

ManBehindMooN:
### Checklist:

- [y] My code follows the [add-on rules](http://kodi.wiki/view/Add-on_rules) and [piracy stance](http://kodi.wiki/view/Official:Forum_rules#Piracy_Policy) of this project. 
- [y] I have read the [CONTRIBUTING](https://github.com/xbmc/repo-plugins/blob/master/CONTRIBUTING.md) document
- [y] Each add-on submission should be a single commit with using the following style: [plugin.video.foo] v1.0.0

Additional information :


plugin.video.srf_ch_replay/addon.py
```diff
@@ -87,12 +80,30 @@ def listTvShows():
...
             picture = ''
-        _addShow(title, show['id'], mode, desc, picture, page)
+
+        firstTitleLetter = title[:1]
+        if (firstTitleLetter.lower() == letter) or (not firstTitleLetter.isalpha() and not str(letter).isalpha()):
+            _add_show(title, show['id'], mode, desc, picture, page)
 
     xbmcplugin.addSortMethod(pluginhandle, 1)
     xbmcplugin.endOfDirectory(pluginhandle)
+    if myView:
```
enen92: Not allowed sorry, plugins can't force skin viewmodes. Please drop this call.

plugin.video.srf_ch_replay/addon.py
```diff
@@ -175,6 +186,9 @@ def listEpisodes(showid, showbackground, page):
     if page < maxpage or maxpage == 0 and len(show) == int(numberOfEpisodesPerPage):
         page = page + 1
         _addnextpage(tr(30005).format(page, maxpage), showid, 'listEpisodes', '', showbackground, page)
+        
+    if myView:
```
enen92: Same

enen92 (changes requested):
Sorry, the add-on is using functionalities that we don't allow

plugin.video.srf_ch_replay/addon.py
ManBehindMooN: @enen92: Is there an allowed way for that? What I wanna basically do is to have the InfoWall ViewType when I'm in the add-on. Right now the default seems to be the list and for every new view or change a page I have to set the view type again. If there is a way could you please refer me to the documentation? Right now this was the only way I have found... Thank you for your help... if there is no way... I will remove it...

enen92: I think there is no other way, the only option is to define a default view for your container content on the skin settings. That applies to all content including add-ons.

ManBehindMooN: @enen92: Is there an allowed way for that? What I wanna basically do is to have the InfoWall ViewType when I'm in the add-on. Right now the default seems to be the list and for every new view or change a page I have to set the view type again. If there is a way could you please refer me to the documentation? Right now this was the only way I have found... Thank you for your help... if there is no way... I will remove it...

ManBehindMooN: ok, then I remove it... thanks for the information

enen92 (approved):

PR #3707, opened 2021-08-25
[plugin.audio.radiorecord] 1.0.1

Nemiroff:
### Description
Update API
### Checklist:

### DIFF
--- a/plugin.audio.radiorecord/addon.xml
+++ b/plugin.audio.radiorecord/addon.xml
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8" standalone="yes"?>
-<addon id="plugin.audio.radiorecord" name="Radio Record" version="1.0.0+matrix.0" provider-name="Nemiroff">
+<addon id="plugin.audio.radiorecord" name="Radio Record" version="1.0.1+matrix.0" provider-name="Nemiroff">
     <requires>
         <import addon="xbmc.python" version="3.0.0" />
         <import addon="script.module.requests" version="2.4.3"/>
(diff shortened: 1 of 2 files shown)


plugin.audio.radiorecord/addon.xml
```diff
@@ -15,13 +15,14 @@
         <disclaimer lang="en_GB">This is not official addon</disclaimer>
         <disclaimer lang="ru_RU">Это не оффициальный аддон</disclaimer>
         <language>ru en</language>
-        <license>GNU General Public License v2.0 only</license>
+        <license>GNU GENERAL PUBLIC LICENSE. Version 2, June 1991</license>
```
enen92: Please use a spdx compliant identifier. In this case: GPL-2.0-only

enen92 (changes requested):
Just a minor

plugin.audio.radiorecord/addon.xml
Nemiroff: Done

enen92 (approved):
Would still be nice if you could use the identifier as the majority of the add-ons do now. Either way, since you don't touch the license line this is good to go

Nemiroff:
> Would still be nice if you could use the identifier as the majority of the add-ons do now. Either way, since you don't touch the license line this is good to go

sorry but i don't understand what about you say

enen92:
In your last commit you left the license string as it was originally. My request when I first reviewed this was for you to use a spdx compliant identifier (GPL-2.0-only) instead of your license string.

Nemiroff:
> In your last commit you left the license string as it was originally. My request when I first reviewed this was for you to use a spdx compliant identifier (GPL-2.0-only) instead of your license string.

use `<license>GPL-2.0-only</license>` ?

enen92:
Yeap. Keep it for the next update 🙂

PR #3760, opened 2021-10-16
[plugin.video.rtpplay] broken

enen92:
Mark addon as broken, supported only for python3 (matrix and above)


enen92:
need https://github.com/xbmc/repo-plugins/pull/3759 to go in first

PR #3804, opened 2021-12-15
[plugin.audio.somafm] 2.0.1

evinr:
### Description
v2.0.1 (2021-12-14)
### Checklist:

Additional information :
Resolves all callouts here: https://github.com/xbmc/repo-plugins/pull/3786 and https://github.com/xbmc/repo-plugins/pull/3798 and https://github.com/xbmc/repo-plugins/pull/3803
Original dev stating intentions for this plugin going forward: https://github.com/Oderik/xbmc-somafm/issues/13#issuecomment-852879925


plugin.audio.somafm/poetry.lock
```diff
@@ -0,0 +1,215 @@
+[[package]]
```
basrieter, line 1: I don't think we need this file?

plugin.audio.somafm/pyproject.toml
```diff
@@ -0,0 +1,22 @@
+[tool.poetry]
```
basrieter, line 1: I don't think we need this file.

plugin.audio.somafm/resources/lib/addon.py
```diff
@@ -0,0 +1,240 @@
...
+
+
+def log(msg):
+    xbmc.log(str(msg), level=xbmc.LOGDEBUG)
+
+
+log(sys.argv)
+
+rootURL = "https://somafm.com/"
+tempdir = xbmcvfs.translatePath("special://home/userdata/addon_data/%s" % __addonid__)
```
basrieter, line 35: This will be down automatically by Kodi I think. No need to do this manually?

plugin.audio.somafm/resources/lib/addon.py
```diff
@@ -0,0 +1,240 @@
...
+                "special://home/addons/%s/fanart.jpg" % __addonid__
+            ),
+        }
+    )
+
+    xbmcplugin.setResolvedUrl(handle, True, list_item)
+
+
+def clearcache():
+    shutil.rmtree(tempdir, True)
```
basrieter, line 218: This will also delete the user's add-on settings. I don't think that this is convenient?

basrieter (changes requested):
Please review my comments

plugin.audio.somafm/resources/lib/addon.py
evinr: Kodi will only create the directory if the settings are changed, otherwise the addon is using the userdata directory to store the channels. Let me know if there is a better place for caching channel data.

PR #3813, opened 2021-12-21
[plugin.video.cbc-sports] 1.0.2

jbinkley60:
Description

Initial new Kodi Matrix addon plugin.video.cbc-sports submission.

Get the latest headlines, opinion, and videos from CBC Sports. Watch live events, replays, and highlights. Some content blocked outside of Canada.

Checklist:

[X ] My code follows the add-on rules and piracy stance of this project.
[X ] I have read the CONTRIBUTING document
[X ] Each add-on submission should be a single commit with using the following style: [script.foo.bar] v1.0.0
[X ] My code follows the add-on rules and piracy stance of this project.
[X ] I have read the CONTRIBUTING document
Each add-on submission should be a single commit with using the following style: [script.foo.bar] 1.0.0

Additional information :

Submitting your add-on to this specific branch makes it available to any Kodi version equal or higher than the branch name with the applicable Kodi dependencies limits.
add-on development wiki page.
Kodi pydocs provide information about the Python API
PEP8 codingstyle which is considered best practise but not mandatory.
This add-on repository has automated code guideline check which could help you improve your coding. You can find the results of these check at Codacy. You can create your own account as well to continuously monitor your python coding before submitting to repo.
Development questions can be asked in the add-on development section on the Kodi forum.


plugin.video.cbc-sports/addon.py
```diff
@@ -0,0 +1,495 @@
...
+from bs4 import BeautifulSoup
+import html5lib
+import json
+from datetime import datetime, timedelta as td
+import time
+
+_addon = xbmcaddon.Addon()
+_addon_path = _addon.getAddonInfo('path')
+selfAddon = xbmcaddon.Addon(id='plugin.video.cbc-sports')
+sys.path.append(os.path.join( _addon_path, 'resources', 'lib' ) )
```
enen92: please avoid using sys.path.append, use relative imports instead:

`from resouces.lib import x`

plugin.video.cbc-sports/addon.py
```diff
@@ -0,0 +1,495 @@
...
+			if (len(splitparams)) == 2:
+				param[splitparams[0]] = splitparams[1]
+
+	return param
+
+
+def addDir(name, url, mode, iconimage, fanart=False, infoLabels=True):
+	u = sys.argv[0] + "?url=" + urllib.parse.quote_plus(url) + "&mode=" + str(mode) + "&name=" + urllib.parse.quote_plus(name) + "&iconimage=" + urllib.parse.quote_plus(iconimage)
+	ok = True
+	liz = xbmcgui.ListItem(name)
```
enen92: you'd get a bit of performance by adding `offscreen=False` in the listitem initializer

plugin.video.cbc-sports/addon.py
```diff
@@ -0,0 +1,495 @@
...
+	if plot:
+	    plot_text = plot
+	else:
+	    plot_text = name
+	aired_text = ''
+	year_test = ''
+	if aired:
+	    aired_text = aired
+	    year_text = aired_text.split('-',1)[0]
+	liz = xbmcgui.ListItem(name)
```
enen92: same here
```suggestion
	liz = xbmcgui.ListItem(name, offscreen=False)
```

plugin.video.cbc-sports/addon.py
```diff
@@ -0,0 +1,495 @@
...
+	CATEGORIES()
+elif mode == 1:
+	INDEX(url)
+elif mode == 2:
+	IFRAME(name,url)
+elif mode == 6:
+	VIDEOS(url)
+elif mode == 7:
+	GET_STREAM(name,url)
+
```
enen92: usually the addon entrypoints are as small as possible and only pass sys.args elsewhere. This is due to the fact the entrypoint is the only file that's not compiled to bytecode.
Please consider do something like this:

```python
import sys
from resources.lib import myplugin

if __name__ == "__ main__":
  myplugin.run(sys.argv)

enen92 (approved):
A few minors if you want to improve, but not mandatory. Let me know.

enen92 (changes requested):
Ups sorry, the sys.path one is mandatory, the others are minors.

plugin.video.cbc-sports/addon.py
jbinkley60: This has been changed to a relative path and sys.path.append has been removed.

plugin.video.cbc-sports/addon.py
jbinkley60: Updated as suggested

plugin.video.cbc-sports/addon.py
jbinkley60: Updated as suggested

jbinkley60:
All suggestions have been implemented

enen92 (approved):

PR #3859, opened 2022-01-22
[plugin.video.hdaustria@matrix] 1.0.5+matrix.1

michaelarnauts:
### Description

- **General**
  - Add-on name: HD Austria
  - Add-on ID: plugin.video.hdaustria
  - Version number: 1.0.5+matrix.1

- **Code location**

Dieses add-on erlaubt Zugriff auf Live TV Kanäle und video-on-demand Inhalte, bereitgestellt von deinem HD Austria Abonnement.

### What's new

v1.0.5 (2022-01-22)
- Fix incorrect time in EPG view.
- Fix hiding of adult channels in EPG view.
- Add translations.

### Checklist:


michaelarnauts:
This Add-on uses the same code as the TV Vlaanderen addon, but it is branded for HD Austria since the API is the same. We've decided to split it up into different add-ons, and build the ZIP with different branding and configuration so it's easier to find in the Kodi repository, this also allows to install the separately. This also makes it easier to manage in case the API's would diverge in the future.

plugin.video.hdaustria/resources/settings.xml
```diff
@@ -0,0 +1,28 @@
+<?xml version="1.0" encoding="utf-8" standalone="yes"?>
```
enen92, line 1: would be nice to get this updated to the new style settings once you have a chance to do it

enen92 (approved):
Looks good, code's pretty similar to other addons from the same authors

plugin.video.hdaustria/resources/settings.xml
michaelarnauts: Good point, I'll take a look at this later.

PR #3861, opened 2022-01-22
[plugin.video.focussat@matrix] 1.0.5+matrix.1

michaelarnauts:
### Description

- **General**
  - Add-on name: Focus Sat
  - Add-on ID: plugin.video.focussat
  - Version number: 1.0.5+matrix.1

- **Code location**

This add-on gives access to the live tv channels and the video-on-demand content available in your Focus Sat subscription.

### What's new

v1.0.5 (2022-01-22)
- Fix incorrect time in EPG view.
- Fix hiding of adult channels in EPG view.
- Add translations.

### Checklist:


michaelarnauts:
This Add-on uses the same code as the TV Vlaanderen addon, but it is branded for HD Austria since the API is the same. We've decided to split it up into different add-ons, and build the ZIP with different branding and configuration so it's easier to find in the Kodi repository, this also allows to install the separately. This also makes it easier to manage in case the API's would diverge in the future.

plugin.video.focussat/resources/settings.xml
```diff
@@ -0,0 +1,28 @@
+<?xml version="1.0" encoding="utf-8" standalone="yes"?>
+<settings>
```
enen92, line 2: it'd be nice if you could move this to the new style settings, it's about time we start to think about deprecating the older alternative

enen92 (approved):

PR #3884, opened 2022-02-07
[plugin.video.lbry] 0.1.9+matrix.1

enen92:
Reverts version 0.1.7 and 0.1.8 of this plugin as the updates were not pushed by the addon author/maintainer. Bumps the version to 0.1.9 to force users to revert to the previous version.

@accumulator please review
@amacchietto please bring your changes upstream if interested, only addon authors/maintainers are allowed to update repo addons unless they are MIA which is not the case

PR #3885, opened 2022-02-07
[plugin.video.lbry] 0.1.9+leia.1

enen92:
Bumps the addon version in Leia to the same revision as the one being pushed to Matrix to avoid confusion. See https://github.com/xbmc/repo-plugins/pull/3884

@accumulator looks like the author did not push the updates to Leia, as a result this doesn't seem to be necessary. Keeping it open in case you want to bump the addon version in Leia to keep it in sync with Matrix. If you don't see a need for this let me know and the PR will be closed. 

Cheers


accumulator:
can also be closed, see https://github.com/xbmc/repo-plugins/pull/3888

PR #3912, opened 2022-03-06
[plugin.program.bluetoothctl] 1.0.1

JimMadge:
### Description

Interact with Bluetooth devices in Kodi using [the Linux Bluetooth stack](http://www.bluez.org/).

### Checklist:


lrusak:
Why not just use the bluez dbus API instead of calling a binary using subprocess?

For an example see how LibreELEC does it. https://github.com/LibreELEC/service.libreelec.settings/blob/master/resources/lib/dbus_bluez.py

JimMadge:
I'm not familiar with D-Bus.

I didn't notice bluez had a D-Bus API. It looks like a better long-term solution, although it would mean adding Python D-Bus bindings as a dependency.

plugin.program.bluetoothctl/addon.py
```diff
@@ -0,0 +1,386 @@
+from functools import wraps
```
pkscout, line 1: It's not good form to have the entire program logic loaded by the file called from the addon.xml (in this case addon.py).  That file in never cached, so there is a significant performance hit.  Instead, import from a different file and just call it from addon.py.

plugin.program.bluetoothctl/resources/lib/bluetoothctl.py
```diff
@@ -0,0 +1,155 @@
+from __future__ import annotations
```
pkscout, line 1: the empty __init__.py file is not needed with Python3 and should be removed.  This also applies for the ones in the resources and lib folders.

plugin.program.bluetoothctl/resources/lib/bluetoothctl.py
```diff
@@ -0,0 +1,155 @@
...
+from typing import Any
+import subprocess
+from subprocess import CompletedProcess
```
pkscout, line 4: Just noting this from the other comment.  Rather than using subprocess and call a binary, you should use the dbus API.

pkscout (changes requested):

PR #3955, opened 2022-04-12
[plugin.video.curiositystream] Fix saving of authentication session.

de-vri-es:
This PR was previously submitted to https://github.com/GMaxera/repo-plugins/pull/2 , but the maintainer does not seem to be available anymore. That is why I am now submitting this here.

This PR fixes session pickling. The file to dump the pickled session to must be opened in binary mode, or it fails with an exception: `TypeError: must be str, not bytes`. This probably broke with python 3. 

Additionally, this PR improves error handling a bit: when unexpected errors occur during authentication, the full exception traceback is added to the event log. Similarly, when unpickling fails other than with a FileNotFoundError, a traceback is logged as warning and an empty session is used.

Finally, it replaces the use of the deprecated `xbmc.translatePath()` with `xbmcvfs.translatePath()`.


basrieter:
Please make the Addon-Checker happy and bump the version of the add-on with each PR

de-vri-es:
Updated the version number and added release notes :)

basrieter (approved):

PR #4047, opened 2022-06-18
[plugin.program.autocompletion@matrix] 2.0.5

finkleandeinhorn:
### Add-on details:

- **General**
  - Add-on name: AutoCompletion for virtual keyboard
  - Add-on ID: plugin.program.autocompletion
  - Version number: 2.0.5

- **Code location**
  - URL: https://github.com/finkleandeinhorn/plugin.program.autocompletion
  
AutoCompletion for the virtual keyboard (needs skin support)

### Checklist:


basrieter:
@sualfred are you fine with this? It seems upstream of this addon is still pointing to phil's version

plugin.program.autocompletion/addon.xml
```diff
@@ -1,18 +1,21 @@
...
     <extension point="xbmc.python.pluginsource" library="plugin.py" />
+
     <extension point="xbmc.addon.metadata">
         <summary lang="en_GB">AutoCompletion for the virtual keyboard (needs skin support)</summary>
         <description lang="en_GB">AutoCompletion for the virtual keyboard (needs skin support)</description>
+        <summary lang="es_ES">Autocompletar para el teclado virtual (necesita soporte de skin)</summary>
+        <description lang="es_ES">Autocompletar para el teclado virtual (necesita soporte de skin)</description>
         <platform>all</platform>
-        <license>GPL-2.0-or-later</license>
+        <license>GNU GENERAL PUBLIC LICENSE Version 2.1, February 1999</license>
```
enen92: This is wrong, should be kept to the SPDX license identifier just like it was before.

plugin.program.autocompletion/addon.xml
```diff
@@ -1,18 +1,21 @@
...
         <source>https://github.com/phil65/plugin.program.autocompletion</source>
```
enen92: if you are the current maintainer this should be changed to point to your repo

enen92 (changes requested):

plugin.program.autocompletion/addon.xml
finkleandeinhorn: As you can see from the commit diff, the original license was `GPL-2.0-or-later`, and the it was changed to `GNU GENERAL PUBLIC LICENSE Version 2.1, February 1999`, which is a compliant license (later than 2.0).

I will change it back anyways, as this is not actually a relevant change to this PR, and the license has indeed not changed.

plugin.program.autocompletion/addon.xml
finkleandeinhorn: I am not intending to continue to maintain this add-on. The last actual update by @phil65 was in 2018, with the latest version being pushed here by @ronie, but I'm not even sure where the original source for 2.0.0 *came from*, as it isn't in the original source repo, and randomly showed up here in 2020. The next update (2.0.1) was from @sualfred, and was only a change in the order of elements in the `addon.xml`. Neither of those updates required changing the maintainer, even though they came years after the original developer seems to have stopped working on it.

There has been a [long-standing issue](https://github.com/phil65/plugin.program.autocompletion/issues/12), ever since Kodi 18 Leia, where this plugin causes Kodi to crash completely.

This PR aims to fix that bug, as well as bring the random features and fixes that are floating around from different developers into a single, accessible codebase. Unless there are other bugs that crop up *from these fixes*, this add-on should not need a new "maintainer" anyways, as it has been largely untouched for the last few years already.

finkleandeinhorn: @enen92 To keep consistent with my other PR, I've gone ahead and updated the `<source>` to my repo here as well.

enen92 (approved):
Maybe this should be changed to the Kodi add-ons gh repo ( https://github.com/XBMC-Addons) at some point.

PR #4104, opened 2022-09-12
[plugin.video.montreal.greek-tv] 4.0.1

greektv:
### Description
Montreal Greek TV is broadcasting in-house locally-produced programming in the Greek language from the Greek Community of Greater Montreal in Canada.
Version 4.0.1

### Checklist:

Additional information :


plugin.video.montreal.greek-tv/resources/lib/default.py
```diff
@@ -33,7 +34,9 @@ def playlinks(url):
         liz = xbmcgui.ListItem(offscreen=True)
         liz.setProperty('isFolder', 'false')
         if url == RADIO_URL :
+            if xbmc.Player().isPlayingVideo() == True : xbmc.executebuiltin("Action(Stop)")
```
romanvm: `is True`

Dimitri-Papadopoulos: it was by inattention that I wrote it this way. I replaced it.
Thank you

plugin.video.montreal.greek-tv/addon.py
```diff
@@ -18,7 +18,9 @@
         along with this program.  If not, see <http://www.gnu.org/licenses/>.
 """
 
-
+import xbmc
 from resources.lib.default import plugin 
 
+xbmc.executebuiltin("InhibitScreensaver(true)")
+xbmc.executebuiltin("InhibitIdleShutdown(true)")
```
romanvm: This is a definite no-go. If you want to disable the screensaver while your addon is running then you can use `try/finally` clause and restore everything as it was in `finally` block. Also an option in your addon setting would be desirable to control this behavior (disabling the screensaver/idle power off).

romanvm (changes requested):

plugin.video.montreal.greek-tv/addon.py
Dimitri-Papadopoulos: I made the change .
 I couldn't find the way  as you suggested but added condition that : inhibit screensaver/idle power off is set to false when no playing streaming. 
Also added the settings so the user can disable  screensaver/idle power off.

greektv:
see please last change

PR #4273, opened 2023-03-05
[plugin.video.rivedila7] V6.2.0

Testato:
6.2.0 (2022-03-5)
- Solved 4h Live delay problem

### Checklist:

### DIFF
--- a/plugin.video.rivedila7/resources/lib/plugin.py
+++ b/plugin.video.rivedila7/resources/lib/plugin.py
@@ -62,7 +62,7 @@ def add_directory_item_nodup(parameters, li, title='', folder=True, is_live=Fals
 
 
 def play_dirette(url, live):
-    xbmc.log('PLAY DIRETTE', xbmc.LOGINFO)
+    # xbmc.log('PLAY DIRETTE', xbmc.LOGINFO)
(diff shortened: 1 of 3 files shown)


plugin.video.rivedila7/resources/lib/plugin.py
```diff
@@ -62,6 +62,7 @@ def add_directory_item_nodup(parameters, li, title='', folder=True, is_live=Fals
 
 
 def play_dirette(url, live):
+    xbmc.log('PLAY DIRETTE', xbmc.LOGINFO)
```
enen92: why loginfo and not logdebug?

Testato: It simple, you do not need enable debug 🤟😅

enen92: That sounds awesome but:

https://kodi.wiki/view/Add-on_rules

addons should use the xbmc.log() method to write to the logfile and use the debug logging level only. print statements should not be used.

Testato: Bureaucrat 😅

enen92 (approved):

PR #4433, opened 2023-12-06
[plugin.video.external.library@nexus] 1.0.0

romanvm:
### Add-on details:

- **General**
  - Add-on name: External Kodi Videolibrary Client
  - Add-on ID: plugin.video.external.library
  - Version number: 1.0.0

- **Code location**
  - URL: https://github.com/romanvm/kodi.external.library
  
A plugin that allows to browse and play items from a video library on an external Kodi instance.

### Description of changes:

1.0.0: Initial public release.

### Checklist:


plugin.video.external.library/libs/mem_storage.py
```diff
@@ -0,0 +1,49 @@
...
+import xbmcgui
+
+
+class MemStorage:
+    """
+    Stores a JSON-serializable Python object in a shared memory within Kodi
+
+    It can be used to exchange data between different Python scripts running inside Kodi.
+    """
+    def __init__(self, window_id=10000):
```
basrieter, line 28: I like the solution, but basically each and every other add-on can access it right? And perhaps overwrite values? 

I would it not make sense to prepend the add-on ID to the keys?

basrieter (approved):
Just a minor question on the memory storage.

plugin.video.external.library/libs/mem_storage.py
romanvm: Yes, the danger storage keys collision exists. I used a rather long key that was supposed to be unique but your suggestion makes sense. I included the addon ID in the key.

PR #4470, opened 2024-02-28
[plugin.video.youtube@nexus] 7.0.3.2

anxdpanic:
### Add-on details:

- **General**
  - Add-on name: YouTube
  - Version number: 7.0.3.2

- **Code location**
  
### Fixed
- Fix settings not being saved and setup wizard repeatedly activating in Kodi 20 and lower
- Fix plugin refresh on settings change

### Checklist:


plugin.video.youtube/resources/lib/youtube_authentication.py
```diff
@@ -7,16 +7,24 @@
     See LICENSES/GPL-2.0-only for more information.
 """
 
-from youtube_plugin.youtube.provider import Provider
-from youtube_plugin.kodion.impl import Context
+from __future__ import absolute_import, division, unicode_literals
```
romanvm, line 10: You don't need this in Python 3.

romanvm (approved):
LGTM

plugin.video.youtube/resources/lib/youtube_authentication.py
MoojMidge: The plugin now has a common codebase for Kodi v18+, was just easier to keep stuff like this in and change the `addon.xml` when making releases, rather than maintaining different branches for different versions of Kodi.

Can remove it for future releases, but would prefer not to, if it is not an issue.

MoojMidge:
@romanvm - not sure if you are the correct person to be directing this to, but there appears to have been an issue with how this submission has been propagated to the addon repo mirror and the Kodi website ‐ the Matrix release file has been used instead of the Nexus release file.

https://mirrors.kodi.tv/addons/matrix/plugin.video.youtube/

https://mirrors.kodi.tv/addons/nexus/plugin.video.youtube/

https://kodi.tv/addons/nexus/plugin.video.youtube/

romanvm:
I will check this ASAP

MoojMidge:
No rush, the Matrix version still works, will just be logging some deprecation notices

MoojMidge:
A related question - is the character used to seperate the revision/version number (`7.0.3.2`) from the build identifier (`matrix.1`) correct?

Currently using `+` but `~` is referred to in https://kodi.wiki/view/Addon.xml#version_attribute, but only for pre-release identifiers.

`+` seems to be used elsewhere eg.
https://mirrors.kodi.tv/addons/leia/service.xbmc.versioncheck/
https://mirrors.kodi.tv/addons/matrix/service.xbmc.versioncheck/
but in this instance the version in the newer target repository has the build identifier, while the older version does not.

This is the opposite to what is currently done for plugin.video.youtube, where the build identifier is used for versions targetting the older plugin repository eg.
`plugin.video.youtube-7.0.3.2` for Kodi 20+
`plugin.video.youtube-7.0.3.2+matrix.1` for Kodi 19

In https://github.com/xbmc/xbmc/blob/40c09d0c9323f7e56a290d06043fbd466de3b915/xbmc/addons/AddonVersion.cpp#L106-L111 the `~` character is specifically checked in the version comparison to sort lower than an empty string, however if `+` is used then this will sort higher, which means that Kodi will try to update a manually installed Kodi 20+ release with the Kodi 19 release if it is available (which normally shouldn't happen, but has in this instance).

So a few questions actually:
1. Which releases should have a build identifier?
1.a) Should releases targetting older Kodi repositories have a build identifier, as currently done in plugin.video.youtube?
1.b) Should releases targetting newer Kodi repositories have a build identifier, as currently done in service.xbmc.versioncheck?
1.c) Should all releases have a build identifer for the minimum compatible Kodi repository version?
2. Should `~` be used instead of `+` for build identifiers in releases targetting older Kodi repositories?
3. Should `+` be used instead of `~` for build identifiers in releases targetting newer Kodi repositories?

MoojMidge:
Settled on the following in discussion with @anxdpanic 
1.c) All releases have a build identifer for the minimum compatible Kodi repository version
3. `+` is used for build identifiers

PR #4509, opened 2024-04-24
[plugin.video.catchuptvandmore@krypton] 0.2.39

joaopa00:

The matrix version was already merged. So this one is for pre-matrix version since some users still use old Catchuptvandmore,  in old Rasperri mainly.
Thanks for your understanding.


joaopa00:
What means the errors?
ERROR: plugin.video.catchuptvandmore addon already exists with a lower or equal version: 0.2.38+matrix.1 in omega branch and the addon has non forward abi compatible dependencies. Users migrating to kodi version omega won't be able to receive the addon update.
ERROR: plugin.video.catchuptvandmore addon already exists with a lower or equal version: 0.2.38+matrix.1 in nexus branch and the addon has non forward abi compatible dependencies. Users migrating to kodi version nexus won't be able to receive the addon update.
ERROR: plugin.video.catchuptvandmore addon already exists with a lower or equal version: 0.2.38+matrix.1 in matrix branch and the addon has non forward abi compatible dependencies. Users migrating to kodi version matrix won't be able to receive the addon update.

basrieter:
It means that the Matrix/Nexus/Omega version of this add-on has a version number lower than 0.2.39, which is not allowed.

basrieter:
Could you create a PR for the Python 3 version of `plugin.video.catchuptvandmore 0.2.39`? 

Because as of today we are no longer accepting Python 2 add-on updates: https://kodi.tv/article/the-sun-sets-on-python-2/

This will then also solve your issue with the versions.

PR #4510, opened 2024-04-27
[plugin.video.catchuptvandmore@matrix] 0.2.39+matrix.1

joaopa00:

Once more, this PR is pretty huge. This was caused by the problem in youtube-dl script on which Catchuptvmore is based. Youtube-dl had to be merged before one can request a PR for Catchuptvandmore

Thanks for your understanding.


basrieter:
Thanks, I will go through the changes this weekend (I hope).

plugin.video.catchuptvandmore/resources/lib/resolver_proxy.py
```diff
@@ -422,7 +508,8 @@ def get_francetv_video_stream(plugin,
         final_video_url = json_parser2['url']
         if download_mode:
             return download.download_video(final_video_url)
-        return final_video_url + '|User-Agent=' + web_utils.get_random_ua()
+        return final_video_url + '|X-Forwarded-For=' + '2.' + str(randint(0, 15)) + '.' + str(
```
basrieter: Why do you need the X-Forwarded-For?

joaopa00: Once, I removed it. Some users complained France did not work anymore. So I added it again. But I always thought something was wrong in their system. You are right.  Better to remove it.

basrieter: I is usually used to circumvent Geo-IP protection. Of course, that is not something we want/allow.

basrieter:
Sorry it is taking some time. Just came back from a week off and away. I am about halfway the PR now.

joaopa00:
Thanks for your feedback. is your review soon finished now?

plugin.video.catchuptvandmore/resources/lib/channels/fr/mycanal.py
```diff
@@ -58,16 +85,128 @@
...
+        return None
+    return base64.b64encode(resp.content).decode('utf-8')
+
+
+def fix_cipher():
+    # Fix a ssl issue on some device.
+    # noinspection PyUnresolvedReferences
+    try:
+        # noinspection PyUnresolvedReferences
+        requests.packages.urllib3.util.ssl_.DEFAULT_CIPHERS += ':HIGH:!DH:!aNULL'
```
basrieter: Please don't do this, as it changes it in all Python environments. Use a custom SSL context: https://github.com/urllib3/urllib3/issues/3173

plugin.video.catchuptvandmore/resources/lib/skeletons/sfrtv_live.py
basrieter, line 1: Just out of curiosity: why is 95% of this file commented?

basrieter (changes requested):
Please have a look at the comments. The 4K lines file with 95% commented, looks odd? And the SSL thing needs to change.

plugin.video.catchuptvandmore/resources/lib/skeletons/sfrtv_live.py
joaopa00: They are channels that need special subscription. I planned to implement them, but I never had the time to do it. So it is better to remove them.

Thanks for the feedback

basrieter (approved):
Thanks for your patience!

joaopa00:
Thank you for your hard work, your many advice to make better this PR and merging this PR.

A one more request please: can you merge  https://github.com/xbmc/repo-resources/pull/476

Thanks in advance

A note. When I forrce-pushed a patch, kodi checker failed. I saw many PR met this problem.  Do I have to worry about that?

PR #4536, opened 2024-07-10
[plugin.video.cbc] 4.0.19+matrix.1

micahg:
### Description

Update to handle recent live TV API changes.

### Checklist:

### DIFF
--- a/plugin.video.cbc/resources/lib/livechannels.py
+++ b/plugin.video.cbc/resources/lib/livechannels.py
@@ -1,4 +1,5 @@
 """Module for live channels."""
+from concurrent import futures
 import json
 from urllib.parse import urlencode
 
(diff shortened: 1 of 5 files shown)


plugin.video.cbc/resources/lib/livechannels.py
```diff
@@ -30,34 +31,59 @@ def get_live_channels(self):
...
             callsign = CBC.get_callsign(channel)
 
             # if the user has omitted this from the list of their channels, don't populate it
-            if callsign in blocked:
+            if f'{callsign}' in blocked:
                 continue
 
             labels = CBC.get_labels(channel)
             image = cbc.get_image(channel)
+
```
basrieter, line 68: I am not completely following this? Is the client key something the user needs to add? Or is the add-on circumventing an authorization mechanic?

micahg: Sorry @basrieter - those notes were added on as I was working through the changes and they are not accurate.

Auth is provided when we start the addon (or when their token expires) and the cookies are preserved on disk.

I'll get rid of these so its not so contradictory!

basrieter (approved):

micahg:
Tyvm @basrieter

PR #4553, opened 2024-08-14
[plugin.video.youtube@matrix] 7.0.9

anxdpanic:
### Add-on details:

- **General**
  - Add-on name: YouTube
  - Version number: 7.0.9

- **Code location**
  
### Fixed
- Fix renaming playlists
- Improve http server wakeup
- Fix ask for video quality setting being inconsistently applied
- Fix http server not sleeping after initial start of service
- Fix not reloading WL playlist if open after playback of video in WL
- Fix navigating to search page after playback and prompt re-opening
- Fix issues with next page and jump to page in related video listings
- Attempt to workaround issue with getting system idle time on Xbox
- Better handle unknown errors in player request responses
- Fix double playback due to busy dialog crash workaround
- Fix disabling certification verification
- Fix not using player request fallbacks
- Fix incorrectly identifying VP9 streams
- Fix videos not able to be re-opened from Kodi playlist
- Fix http server not working in Kodi 18
- Fix issues with http server wakeup and sleep on initial start/restart
- Fix API requests failing if not logged in
- Fix multi-audio streams failing to play
- Fix not properly defining default audio role
- Don't retry player requests for offline live streams
- Fix incorrectly updating playing live stream details
- Fix typo causing http server to restart when checking status

### Changed
- Make live query parameter optional when playing channel live stream
- Allow for removal of custom watch later and history playlists from settings
  - Use default value of "HL" for history or "WL" for watch later as reset value
  - This will then clear the setting and clear value from ac

### Checklist:


basrieter (commented):
I will continue checking later. So far OK

plugin.video.youtube/resources/lib/youtube_plugin/youtube/client/youtube.py
```diff
@@ -1443,167 +1391,386 @@ def search(self,
...
+                'kwargs': channel_ids,
+                'output': feeds,
+                'worker': _get_feed,
+                'threads': threads,
+                'limit': None,
+                'dynamic': True,
+                'input_wait': threading.Lock(),
+            },
+        ))
+        while 1:
```
basrieter, line 1711: Why the need for a `while 1`? It seems like this might cause a lot of threads to spawn?

basrieter (changes requested):
Please double check my comment.

plugin.video.youtube/resources/lib/youtube_plugin/youtube/client/youtube.py
MoojMidge: The number of threads that can be created is defined in line 1667-1669

This is checked in lines 1726-1734 to allow threads in the pools to be re-balanced (i.e. to be shutdown/repurposed and released allowing new threads to be added to the pool if needed)

The loop execution is also being paused if the number of threads exceeds the max allowable through the use of a threading.Semaphore object (created line 1670, acquired line 1744, and released by the various threads that are spawned on line 1657).

This means that the creation of new threads in the loop is always constrained, but still allows for a variable amount of threads to be monitored within the loop, spawning new threads if required, but also pausing thread execution if it has nothing to do, skipping idle threads while waiting for input from other threads, and releasing threads if their job has finished. 

Edit: the previous implementation used a simple for loop, which meant the number of threads were constrained by the number of items being looped over. Unfortunately these elements are the individual channels that someone may have subscribed to, which meant that the previous implementation was potentially spawning hundred/thousands of threads all at once in an effort to fetch each channel in parallel.

basrieter (approved):

basrieter:
Thanks for your patience. Perhaps for the future: This PR was really huge. I understand that sometimes you cannot prevent this. But multiple smaller PR's go through the review faster than a singe huge one.  Perhaps something to keep in mind for future PR's.

MoojMidge:
Well noted.

Unfortunate circumstance of having to rush out the previous release to beat the previous round of YouTube changes, and subsequent releases adding features while simultaneously trying to keep up with even more YouTube changes, that has only now settled down (until it all starts again).

I'll try to organise things a bit better for future submissions to try and reduce the number of changes coming in one go.

Issue #4568, opened 2024-09-16
Radio.de add is not working anymore


georgkrause:
### Add-on details:

- Add-on name: Radio
- Add-on ID: plugin.audio.radio_de
- Version number: 3.0.9+matrix.1
- Kodi/repository version: 21/3.4.0

### Ownership
- [ ] I am the owner of this add-on

    
### Broken because:

Whatever action I want to take in the addon, there is a network error. I think the radio.de API changes and the changes are not yet reflected in the repository.

I am not aware if thats the right ways to report such issue.  In the addon-repository the malfunction is already reported, see https://github.com/XBMC-Addons/plugin.audio.radio_de/issues/61 or https://github.com/XBMC-Addons/plugin.audio.radio_de/issues/59 . However, no action is taken. I'd suggest to remove the addon from the official repository until its fixed.

If you are aware of any maintainer of the addon, I might be interested in helping getting the plugin running again. Let me know if there is a way to get involved.

basrieter:
The right place to report this, is at the source repository of the add-on. Not here.

georgkrause:
@basrieter I report it here because as I mentioned, its already reported in the source repository. But this belongs to the same org as the index, however, nobody replies there. As I already wrote in the issue, I am asking you to remove a non-functional addon from the repository to avoid people getting annoyed by installing a broken plugin.

georgkrause:
@basrieter I am especially wondering why there is a category to report broken addons if you don't want those issues to be opened. Am I missing something?

basrieter:
The original add-on author needs to create e PR to mark an add-on as `broken`, if and only if he/she does not intent to fix it. A broken add-on will be removed on all Kodi installs.

Only of the owner is unknown or not active we could do this. So let's just wait for now and I will reach out to the devs in the meantime.

georgkrause:
@basrieter Thank you for the info! I filed a PR to mark the addon as broken, before I've read your comment. Sorry if this is useless now, but I think its fair to mark the addon as broken for now.

Just so you are aware: I am planning into fixing the issue, but it might take a while.

basrieter:
No worries, I will tag the original devs for an OK.

PR #4569, opened 2024-09-16
Mark radio.de addon as broken

georgkrause:
### Description

As reported in #4568 the addon is not working anymore. The issue was reported to the projects issue tracker already, however its months now and there wasn't any reaction. I expect the plugin to be broken for a while, even though I actually plan to debug and eventually fix the issue. Until this is done, its only fair not to let people install the addon if its not working anyways. 

I am not sure if I have to bump the version in order to mark it as broken. If there is something missing, let me know and I am happy to touch up!

### Checklist:

### DIFF
--- a/plugin.audio.radio_de/addon.xml
+++ b/plugin.audio.radio_de/addon.xml
@@ -1,4 +1,4 @@
-<addon id="plugin.audio.radio_de" name="Radio" provider-name="Tristan Fischer, enen92" version="3.0.9+matrix.1">
+<addon id="plugin.audio.radio_de" name="Radio" provider-name="Tristan Fischer, enen92" version="3.0.9+matrix.2">
     <requires>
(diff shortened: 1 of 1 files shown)


basrieter:
@gade01 @yol @anxdpanic @MartijnKaijser Are you guys OK with this?

georgkrause:
It looks like I have to bump the version, I'd say the correct one would be `v3.0.9+matrix.2`?

basrieter:
> It looks like I have to bump the version, I'd say the correct one would be `v3.0.9+matrix.2`?

Indeed. Or make it `v3.0.10+matrix.1`.

georgkrause:
But if the plugin eventually releases v3.0.10 we would be in trouble. I therefore decided for v3.0.9-matrix.2

basrieter:
Then they need to bump it to `v3.0.11`, so no worries.

PR #4596, opened 2024-11-10
[plugin.switchback@nexus] 1.0.0

bossanova808:
### Add-on details:

- **General**
  - Add-on name: Switchback
  - Add-on ID: plugin.switchback
  - Version number: 1.0.0

- **Code location**
  - URL: https://github.com/bossanova808/plugin.switchback/
  

Kodi utility for fast switching between recently played media.

Keeps a list of recently played media - supports Kodi library episodes, movies, PVR channels, songs and non-library files.  Does not (yet?) support add-ons or PVR recordings.

The primary intended use is to make for super easy Switchback between two in progress videos (by binding a remote key to Switchback, see Wiki/Forum thread for info).

Consider this scenario:
- You are watching 'video A' with your partner.  You are interrupted and your partner needs to tend to the kids/howl at the moon.
- You navigate to 'video B' and watch some of that.
- Your partner comes back.
- You hit your one button 'Switchback' and 'video A' starts playing again, no need for tedious navigation etc.
- You are interrupted again - the moon is so very bright tonight - hit your 'Switchback' to resume 'video B' immediately, again with one magic button and no tedious navigation.
        

### Description of changes:

            v1.0.0 Initial release
        

### Checklist:


bossanova808:
@basrieter As requested, now a PR to Nexus branch.

plugin.switchback/resources/lib/playback.py
```diff
@@ -0,0 +1,35 @@
...
+    duration:float = None
+    artist:str = None
+    album:str = None
+    tracknumber:int = None
+    channelname: str = None
+    channelnumberlabel:str = None
+    channelgroup:str = None
+
+    def toJson(self):
+        return json.dumps(self, default=lambda o: o.__dict__)
```
basrieter: I would just return a dictionary here and do the dumping to string in the place where you save stuff.

plugin.switchback/resources/lib/store.py
```diff
@@ -0,0 +1,95 @@
...
+        Logger.info(Store.switchback_list)
+
+    @staticmethod
+    def save_switchback_list():
+        """
+        Save the Switchback list to file in JSON format
+        :return:
+        """
+        with open(Store.switchback_list_file, 'w', encoding='utf-8') as f:
+            temp_json = []
```
basrieter: Use my previous comment and let `toJson` return a dictionary and then just do a `json.dumps` of the full array here.

plugin.switchback/resources/lib/switchback_service.py
```diff
@@ -0,0 +1,31 @@
...
+
+    while not Store.kodi_event_monitor.abortRequested():
+        # Abort was requested while waiting. We should exit
+        if Store.kodi_event_monitor.waitForAbort(1):
+            break
+        # Otherwise, if we're playing something, record where we are up to, for later resumes
+        # (Playback record is created onAVStarted in player.py, so check here that it is available)
+        elif Store.current_playback and Store.kodi_player.isPlaying():
+            Store.current_playback.resumetime = Store.kodi_player.getTime()
+            xbmc.sleep(500)
```
basrieter: Use `Store.kodi_event_monitor.waitForAbort` here.

basrieter (changes requested):
Please have a look at my comments.

bossanova808:
I think that is all addressed.  

(I'm not really super clear on the `waitForAbort` vs `sleep` behaviour but let me know if this isn't right!)

basrieter (approved):

basrieter:
Thanks for you contribution!

PR #4598, opened 2024-11-28
[plugin.audio.radiofrance] v1.1.4

bateast:
### Description
Robustify + speedup through http request caching and multithreading

### Checklist:

### DIFF
--- a/plugin.audio.radiofrance/default.py
+++ b/plugin.audio.radiofrance/default.py
@@ -1,162 +1,115 @@
+import os
+import datetime
 import json
+
 import sys
 import requests
+from pathlib import Path
 from urllib.parse import parse_qs
-from enum import Enum
+from concurrent.futures import ThreadPoolExecutor
+import itertools
 
-# http://mirrors.kodi.tv/docs/python-docs/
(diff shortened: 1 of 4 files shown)


plugin.audio.radiofrance/default.py
```diff
@@ -1,163 +1,112 @@
...
 import xbmcgui
 import xbmcplugin
 
 from utils import *
+from interface import *
 
 DEFAULT_MANIFESTATION = 0
 RADIOFRANCE_PAGE = "https://www.radiofrance.fr"
 
+CACHE_FILE = os.path.join(xbmcvfs.translatePath('special://temp/'), 'my_cache.json')
```
basrieter: I would use a name that is unique to your add-on, as `my_cache.json` does not really identify that it belongs to your add-on and might cause issues in the future if another add-on uses the same name.

basrieter (changes requested):

plugin.audio.radiofrance/default.py
romanvm: I'd recommentd to store all addon-specific data including cache in the addon profile dir:
`Addon().getAddonInfo('profile')`

basrieter (approved):

PR #4603, opened 2024-12-14
[plugin.video.youtube@matrix] 7.1.1.6

anxdpanic:
### Add-on details:

- **General**
  - Add-on name: YouTube
  - Version number: 7.1.1.6

- **Code location**
  
### Fixed
- Fix http server not listening on any interface if listen IP is 0.0.0.0
- Standardise return type of LoginClient.refresh_token
- Fix curl headers not being used when set on path of setResolvedUrl listitem
- Fix HEAD requests to MPD manifests
- Fix various Python2 incompatible changes
- Properly distinguish between VP9 and VP9.2 with HDR info
- Fix http server not running when script shows client IP
- Fix not listing full stream details in selection dialog
- Fix translated subtitles not being available for some videos
- Fix labelling for auto dubbed audio tracks
- Fix regressions in handling VP9.2 video streams
- Fix error handling items without headers
- Fix incorrectly determining Kodi release name
- Fix search window history navigation when using direct links
- Fix Python2/Android incompatibilty when checking CPU count
- Fix method used to determine address of local network interface
- Fix potential infinite loop with corrupted access_manager.json
- Various improvements to multiple busy dialog crash workarounds
- Fallback to search listing when search query is empty
- Fix debug log string formatting error
- Fix duplicated separators in context menu
- Fix Kodi builtin using wrong playlist type
- Ensure http server is started prior to creating MPD for playback
- Fix shuffle play of playlists not working
- Fix plugin category label not being applied if content type not set
- Fix Kodi navigating to root path of Video window if plugin listing fails
- Fix l

### Checklist:


plugin.video.youtube/resources/lib/youtube_plugin/youtube/helper/stream_info.py
```diff
@@ -1373,111 +1396,138 @@ def load_stream_info(self, video_id):
...
+                        )
                 else:
-                    self._context.log_debug(
-                        'Unknown playabilityStatus in player response:\n|{0}|'
-                        .format(playability)
-                    )
+                    break
+                if not restart:
+                    break
+                restart = False
```
basrieter, line 1504: Should you not wait a bit here?

basrieter (changes requested):
Please have a look at remark?

plugin.video.youtube/resources/lib/youtube_plugin/youtube/helper/stream_info.py
MoojMidge: Yes, maybe, but not exactly, I think :)

Yes - it's probably good practice.

Maybe - it's a service that can easily handle the requests, and other 3rd party clients and even the official clients perform similar requests without waiting.

But not exactly - a wait time, if used, would probably be better placed in the existing `for` loop that makes the requests, rather than in the new `while` loop, that was added to reset the existing `for` loop.

The master branch that is pushed to this repository has already been updated, and it is a bit of a pain to reset it back to make a non-conflicting commit. Can do so if necessary, but otherwise can instead add this in the next version?

MoojMidge: Actually on second thought I don't think there should be any wait time between requests.

I can fill the connection table of my local router with parallel requests (numbering in the thousands) without seeing any issues with remote responses. There are defined error responses for too many requests but i can't seem to hit that limit, so the 2-4 requests made sequentially here shouldn't be a problem.

Given that the official clients do something similar but using QUIC, I think the logic is that it is less resource intensive to respond to multiple requests using the same connection rather than waiting and potentially opening new connections.

basrieter: I don't really care about the service not being able to handle, more that if the loop keeps going, it might cause Kodi to freeze?

MoojMidge: Well if the loop keeps going it will probably make Kodi unresponsive regardless of any time delay that was added.

I don't believe the loop can keep going because of the different way in which the service responds to unauthenticated vs authenticated requests.

Just in case the request responses do change in the future, I will add something like the following to the next release.

```
Subject: [PATCH] Avoid potential player request loop in case future request responses do not distinguish between (un)authenticated requests
--- a/resources/lib/youtube_plugin/youtube/helper/stream_info.py
+++ b/resources/lib/youtube_plugin/youtube/helper/stream_info.py
@@ -1414,7 +1414,7 @@ class StreamInfo(YouTubeRequestClient):
             if name == 'ask' and use_mpd and not ask_for_quality:
                 continue
 
-            restart = False
+            restart = None
             while 1:
                 for client_name in clients:
                     _client = self.build_client(client_name, client_data)
@@ -1480,7 +1480,10 @@ class StreamInfo(YouTubeRequestClient):
                         )
                         compare_reason = _reason.lower()
                         if any(why in compare_reason for why in reauth_reasons):
-                            if has_access_token:
+                            if client_data.get('_auth_required'):
+                                restart = False
+                                abort = True
+                            elif restart is None and has_access_token:
                                 client_data['_auth_required'] = True
                                 restart = True
                             break
```

MoojMidge:
Thanks for going through this

PR #4645, opened 2025-03-18
[plugin.video.tv3.cat] Delete as it is outdated

PereRB-GH:
### Description
Requesting to delete this plugin as the provider has migrated to another domain with another URLs. I've made this [plugin](https://github.com/xbmc/repo-plugins/pull/4643) to follow the stream. 

This plugin is using URLs like [http://www.ccma.cat/tv3/alacarta/apm/](http://www.ccma.cat/tv3/alacarta/apm/), which you can see it's a 404 not found. 

Have not find any documentation as how to request decommissioning of outdated plugins. 

### Checklist:


basrieter:
There first needs to be a PR (with a new version) that marks it as broken. That will cause it to be disabled on all Kodi version.

PereRB-GH:
Maybe I can upgrade it to work with the new website... For now thinking/working on it. Can you mighty maintainers confirm that there is no need for involvement of the original developer?

romanvm:
If the original developer abandoned the project then I see no problem. It would be polite to ask the original dev if it's ok to take over the project  but if it's not possible or the original dev is not responsive then you are free to do whatever you want within the license terms.

PereRB-GH:
Deleting this PR and will create a new one based on the given instructions.

PR #4697, opened 2025-08-04
[plugin.video.youtube@matrix] 7.2.0.3

anxdpanic:
### Add-on details:

- **General**
  - Add-on name: YouTube
  - Version number: 7.2.0.3

- **Code location**
  
### Fixed
- Fix performance regression with channel filters
- Fix regression preventing video playback from Kodi Info dialog
- Fix exception in youtube.helper.v3.response_to_items when using Kodi 18 (Python 2)
- Fix exception in YouTubeRequestClient.json_traverse when using Kodi 18 (Python 2)
- Fix handling of custom Watch Later playlist in add to playlist dialog
- Fix Liked video playlist not loading
- Fix typo preventing some video only searches from working
- Fix incorrect deletion of provider parameter in 10c3758
- Fix possible hangs in multiple busy dialog workaround
- Fix misidentifying channels in channel filter
- Attempt to avoid race condition where user input occurs prior to window rerouting
- Fix incorrectly referencing channels as playlist in My Subscriptions after db44928
- Improve handling of invalid playlists in My Subscriptions after db44928
- Update YouTube.get_related_videos and YouTube.get_related_for_home
- Fix error with listing comments if number of replies is 1000+
- Fix regression in opening playlists of logged in user
- Fix regressions in checking channel filters
- Clear old function cache values on initial run after install
- Fix My Subscriptions not loading old or previously accessed content
- Manually redirect video server requests if required
- Improve logging of client ID
- Fix possible exception that may occur if video thumbnails fail to update
- Fix unsubscribing from My Subscriptions
- Fix folders for channel shorts and live stream

### Checklist:


basrieter (approved):
Looks ok. 

Would it be possible to reduce the size of future PRs? Reviewing changes of this magnitude is extremely time-consuming, and with so many modifications bundled together it’s nearly impossible to properly assess the impact of each change.

Instead of submitting very large PRs every six months or so (as has been the pattern for this plugin), could you aim for smaller releases, perhaps monthly? Smaller PRs are generally picked up faster, are easier to comprehend, and make the review process far less time-consuming.

MoojMidge:
Can certainly try, but it may not make things any easier.

The releases for this addon have actually been coming out every 1-2 months, just not being pushed to the official repository.

Unfortunately the monthly releases are not really that much smaller than the releases that make it to here, as they tend to be snapshots of the longer term work in updating the framework that the addon was built upon, while also trying to keep up with an ever changing content provider and adding new functionality.

Trying to balance keeping the addon working while getting enough testing done on the framework changes means that there tends to be multiple code paths for the same/similar functionality in the more regular releases, which then eventually gets consolidated in the releases that make it to here.

If it helps any in the overall time needed for reviews, willing to look at any other addon PR reviews that you want to offload.

basrieter:
I understand. Just keep it in mind for the future. Smaller PR's usually pass the reviews faster.

PR #4725, opened 2025-11-30
[plugin.video.youtube@matrix] 7.3.0

anxdpanic:
### Add-on details:

- **General**
  - Add-on name: YouTube
  - Version number: 7.3.0

- **Code location**
  
### Fixed
- Revert change to busy polling interval
- Prune invalid entries from DB when closing connection
- Fix regressions with SQLite db operations
- Disable label masks being used in Kodi 18
- Python 2 compatibility workaround for lack of timeout when trying to acquire an RLock
- More expansive handling of inconsistent urllib3 exception re-raising
- Fix regression in handling audio only setting after d154325c5b672dccc6a17413063cfdeb32256ffd
- Fix comments not using correct sort methods
- Fix incorrectly using playlist cache entries that have been invalidated by playlist modification
- Fix some context menu actions failing for video item bookmarks
- Ensure listings and items added by the addon have correct sort order
- Fix resetting client region when playing media with subtitles enabled
- Only add playable items to playlist when adding related items
- Fix using invalid default end limit with Playlist.GetItems JSONRPC method
- Fix conversion of SRT subtitles to WebVTT
- Workaround playback failure of progressive streams
- Fix re-sorting live search lists
- Disable use of custom thumbnail urls
- Workaround addon service not starting prior to plugin invocation
- Fix unofficial version using localised sort order
- Fix parsing of logged_in query parameter
- Fix typo in YouTubePlayerClient error hook
- Fix not resolving single playable items when using the uri2addon plugin endpoint
- Correctly check whether access tokens are available to be used for player requests
- 

### Checklist:


MoojMidge:
Another large release, hopefully the last as I move towards making changes in more manageable interim releases.

plugin.video.youtube/resources/lib/youtube_plugin/kodion/plugin_runner.py
```diff
@@ -20,70 +27,115 @@
...
+                     print_callees=False,
+                     num_lines=20)
 
 
 def run(context=_context,
+        log=_log,
         plugin=_plugin,
         provider=_provider,
         profiler=_profiler):
+    gc.disable()
```
basrieter: Could you make sure that there is a `try-finally` here, where in finally `gc.enable()` and `gc.collect()` is always called. Kodi re-uses Python instances and if your add-on fails, then the gc is disabled for that re-usable instance.

basrieter (commented):
Ok, it took me a lot of time. One thing I would like to see changed, and that is the `gc.disable()` without a proper `finally` that re-enables it.

basrieter (changes requested):
See my previous comment.

plugin.video.youtube/resources/lib/youtube_plugin/kodion/plugin_runner.py
MoojMidge: > Kodi re-uses Python instances and if your add-on fails, then the gc is disabled for that re-usable instance.

With hindsight I think I should not be trying to do this.

While it speeds up long running plugin calls by a significant amount, I think this actually influences all subinterpreters not just the one used by this addon.

While I test the addon using timing code that implicitly disables the gc anyway, can't be certain that there aren't any cyclic object references that may result in memory leaks.

A better way may be to use a try/finally to prevent automatic collection then collect manually at the end of the plugin call - most of the speed up with less of the unknown potential issues

MoojMidge:
Addon submitter force pushes changes to existing PRs.

Specific changes from what you originally looked at:
https://github.com/anxdpanic/plugin.video.youtube/compare/v7.3.0...v7.3.0.1

basrieter (approved):

PR #4753, opened 2026-03-04
[plugin.video.youtube@matrix] 7.4.1

anxdpanic:
### Add-on details:

- **General**
  - Add-on name: YouTube
  - Version number: 7.4.1

- **Code location**
  
### Fixed
- Fix unnecessary double resolve on playback
- Attempt to break busy dialog workaround loop if post run action succeeds
- Don't swallow unhandled plugin exceptions without logging and running listing end callback
- Python 2 compatibility fix for urllib3 only checking for str type when accessing module constants using getattr
- Always refresh missing access tokens if partially logged in and refresh tokens are available
- Fix missing post run plugin runner option used when search input is a plugin url
- Fix incorrectly identifying and removing Opus audio from list of available streams
- Clean up window properties in case of exception during plugin invocation

### Changed
- Simplify caching and fetching of updated subscriptions for My Subscriptions
- Always load cached My Subscriptions feed data and load cached data if feed update fails
- Use available access token when first initialising client rather than setting access tokens at later stage of client initialisation
- Don't cache main plugin menu so that Sign-In/Out item always reflects actual logged in status
- Automatically add/remove items from local watch later list when using YouTube WL list

### New
- Add Auto-like after watch feature with optional channel filtering

### Checklist:


MoojMidge:
@kodiai - can you please clarify what you mean by the following statement?

> Python's `set.pop()` is not thread-safe - concurrent calls can cause RuntimeError, data corruption, or missing items when threads race to pop from the same set.

Are you indicating that the underlying CPython implementation of set.pop() is not atomic, or that multiple threads can concurrently pop the same or different elements from the same shared set?

This does not seem to be what is documented for GIL enabled versions of CPython, nor what is documented for versions where free-threading is enabled, nor what the code for the set.pop() seems to indicate (https://github.com/python/cpython/blob/main/Objects/setobject.c#L901-L902).

Are you instead try to say that concurrent set.pop() operations may possibly result in attempting to pop an element from a empty shared set, which will result in a KeyError exception being raised, but would not result in the "RuntimeError, data corruption, or missing items" that you have described?

MoojMidge:
@kodiai - the possible race condition is now handled, please revise/review again.

romanvm (approved):

PR #4767, opened 2026-04-06
[plugin.video.youtube@matrix] 7.4.2

anxdpanic:
### Add-on details:

- **General**
  - Add-on name: YouTube
  - Version number: 7.4.2

- **Code location**
  
### Fixed
- Ensure updated context is used in client instance when rerouting to existing provider handler methods
- Fix required ID potentially not being passed to YouTubeDataClient.(add|remove)_video_from_playlist()
- Skip unplayable live manifests
- Handle inconsistent error codes and names used in Python 2
- Simplify byte range to timestamp conversion for compatibility with Python 2
- Fix unnecessarily reloading listing when adding items to Bookmarks

### Changed
- Update parsing of plugin urls for path params and ID lists
- Rework addon debug log settings (Addon > Settings > Advanced > Debug > Enable debug logging)
  - Allowable values:
    - Disabled (addon logging output disabled)
    - Auto (default, addon logging output enabled if Kodi debug logging enabled)
    - Always enabled (addon logging output enabled)
    - Verbose (additional addon logging output enabled)
- Various other improvement to addon logging

### New
- Update Setup Wizard to account for user possibly using mismatched watch history settings

### Checklist:


MoojMidge:
@kodiai - Why have you reviewed this submission twice and reported different findings in each instance? Please also clarify what you mean by the following:

> While the code re-raises after filtering known errors, this prevents proper error reporting for unexpected exceptions between the catch and the errorcode check. If an exception occurs that is not HTTPError and doesn't have an errno attribute matching SWALLOWED_ERRORS, it gets logged but execution continues instead of propagating the original exception type.

This is incorrect, execution does not continue. Any "unexpected exceptions" are re-raised as you have already identified.

> Masks unexpected errors between the catch and errorcode check, potentially hiding serious bugs or security issues.

What "serious bugs", "security issues" or otherwise "unexpected errors" are being masked or hidden here?

> `context_changed()` uses `self._context != context` which compares object identity. If context is reconstructed with identical values, this incorrectly reports a change, causing unnecessary client recreation in provider.py:1806. Should implement proper equality comparison or check relevant context attributes.

Why do you believe it is incorrect to report that a different context instance being used, is a change that requires the client instance to be recreated?

What makes you think that a specifically cloned context instance, that can change compared to the original context instance, should not result in the client instance being re-created or re-initialised?

In any case, why specifically is "unnecessary" client object (re-)creation a "MAJOR" issue that should prevent this PR from being merged?

> Uses object identity (`!=`) instead of value equality for context comparison. Causes unnecessary client recreation when context objects are cloned with identical values, leading to performance degradation and potential loss of cached state.

Please quantify what performance degradation will result from creating a new object instance here and why this is a "MAJOR" issue.

Please clarify what potential loss of cached state would result, and why you think this is a "MAJOR" issue as opposed to the intended purpose of this change?

romanvm (approved):

PR #4791, opened 2026-07-15
[plugin.video.channelme] 1.1.0

GaianHelmers:
### Description - Channel Me! v1.1.0
Tackling all the lingering feature requests, needs should drastically slow down now. Features are fairly exhaustive!
- Per-show bias: give any title its own back-to-back run - a channel default (Up to N,
  or Always N for a fixed run) plus a per-show override - so a dense show can run in
  longer blocks among smaller ones.
- Selection weight: bias how often each show is picked (1.0x - 3.0x) so a big library
  shows up proportionally more without exhausting your smaller shows.
- Sleep timer: optionally stop playback after a set time (5 / 7 / 10 / 15 / 30 / 45 /
  60 / 90 / 120 minutes) with an on-screen countdown you can Cancel or Restart before
  it fires.
- Don't change watched status: optionally play channels without touching your library's
  watched flags, play counts or resume points, even when stopping mid-episode (on by
  default).
- No-repeat shuffle: in pure random, an episode no longer repeats until the rest of that
  show has played.
- "The" sorting: editor title lists follow Kodi's own "ignore The when sorting" setting.
- Editor: back-to-back and selection weight are now numeric sliders; added an add-on
  settings screen for the sleep timer and watched-status options.
  
### Checklist:


GaianHelmers:
I am guessing some of the more problematic markers are things like modifying the playback state of videos played through the Channels.

The system records the playback state: unwatched, ongoing, watched, as well as the playcount count and preserves it after channel viewing, including when interrupting playback.

This emulates the same sort of feature implemented in another random playback addon, and is really helpful when you're just arbitrarily watching shows and don't want it to actually affect your Shows and Movies front page, or leave a bunch of interrupted shows in random parts of your library.

---

For the benefit of the configuration I also integrated a custom slider element for better configuration of channel settings.

While larger in the interface, it complies with the behaviour of standard Kodi sliders:
- You navigate to it to highlight.
- You select to edit the actual value.
- You select to exit the edit
- You navigate freely thereafter.

It displays 3 states: Unhovered, Hovered, and Focused/editing. Visual tinting was a pain and not working, so I opted to just use different graphics, one for each colour state: dim, white, and active-blue.

---

A new Settings page is used to define the "dont affect playcount" toggle, sleep timer, and timer warning prompt times.

---

If desired, I can update the screenshots for the edited windows:
- Mode
- Back to Back
- Settings.

I'm just tired and don't have a good way to get the screenshots.

GaianHelmers:
I got some! (Updated my local Repo Readme: https://github.com/GaianHelmers/plugin.video.channelme/)

Back-to-Back

mode

Settings

PR #4803, opened 2026-08-07
[plugin.switchback@nexus] 2.0.1

bossanova808:
### Add-on details:

- **General**
  - Add-on name: Switchback
  - Add-on ID: plugin.switchback
  - Version number: 2.0.1

- **Code location**
  - URL: https://github.com/bossanova808/plugin.switchback/
  

Kodi utility for fast switching between recently played media.
Keeps a list of recently played videos.
The primary intended use is to make for super easy Switchback between two in progress video. See Wiki for usage notes including how to bind remote/keyboard controls for instant Switchback.
Consider this scenario:
* You're watching 'video A' with your partner. You're interrupted, and your partner needs to tend to the kids/howl at the moon.
* You navigate to 'video B' and watch some of that while you're waiting.
* Your partner comes back.
* You hit your one button 'Switchback' and 'video A' instantly starts playing again - no need for tedious navigation etc.
* You're interrupted, again! The moon really is so very bright tonight. Hit your 'Switchback' button to instantly resume 'video B', again with no tedious navigation.
        

### Description of changes:

v2.0.1
- Playback tracking now uses a shared class from script.module.bossanova808, common with other addons
- Quick Switchback now resumes PVR live TV and recordings correctly, with proper channel controls
- On Kodi 22 (Piers) and newer, PVR live TV and recordings now resolve correctly with no workaround needed at all - including full channel controls for the Switchback list too, not just Quick Switchback - thanks to a Kodi core fix (https://github.com/xbmc/xbmc/pull/28893, fixing https://github.com/xbmc/xbmc/issues/28877). Older Kodi versions keep the previous behaviour.
- Fix a rare crash when saving the Switchback list on Windows
- Resume notifications now show a timestamp, distinguish live TV/recordings from other video, and note that retuning live TV may take a few seconds
- Fix duplicate list entries appearing for the same movie/episode/etc across repeated plays (could happen for addon-backed library items, e.g. Jellyfin, where the reported path can differ between plays)
- Faster list loading: halved the JSON-RPC calls needed to refresh a list of library items, and removed an unnecessary call on every list view
        

### Checklist:

### DIFF
--- a/plugin.switchback/addon.xml
+++ b/plugin.switchback/addon.xml
@@ -1,9 +1,8 @@
 <?xml version="1.0" encoding="UTF-8" standalone="yes"?>
-<addon id="plugin.switchback" name="Switchback" version="2.0.0" provider-name="bossanova808">
+<addon id="plugin.switchback" name="Switchback" version="2.0.1" provider-name="bossanova808">
     <requires>
-        <import addon="xbmc.python" version="3.0.0" />
-        <import addon="script.module.bossanova808" version="1.0.2" />
-        <import addon="script.module.infotagger" version="0.0.7" />
+        <import addon="xbmc.python" version="3.0.1" />
+        <import addon="script.module.bossanova808" version="1.0.4" />
     </requires>
     <!-- The main service entry point, to keep track of playbacks -->
     <extension point="xbmc.service" library="service.py" />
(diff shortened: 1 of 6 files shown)


bossanova808:
(NB Depends on https://github.com/xbmc/repo-scripts/pull/2877 going in first)

bossanova808:
Closing this - I've retargeted Switchback's submission to the matrix branch instead of nexus (matrix is our normal baseline for this addon; nexus was only ever needed because of a dependency that's since been removed). See #4805.

PR #4804, opened 2026-08-08
[plugin.video.spursplay] v0.3.0

LS80:
### Description

Update to fix broken login and make live streams pausable:

- **Fix broken login**: SPURSPLAY moved sign-in to Auth0 Universal Login, so the old direct login endpoint no longer worked at all. Also fixes a bug where the login token failed to persist to settings, forcing a fresh credential login on every screen navigation and triggering CAPTCHA blocks from the auth provider.
- **Make live content pausable**: live streams are now routed through inputstream.adaptive (declared as an optional dependency) instead of Kodi's built-in demuxer, which treated live HLS as non-seekable. Pause and limited seek within the live DVR window now work.

### Checklist:

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/plugin.video.spursplay/addon.py
+++ b/plugin.video.spursplay/addon.py
@@ -41,24 +41,59 @@ def build_url(**params):
 
 def token_expiry(jwt):
     payload_base64 = (jwt.split(".")[1] + "==").encode("ascii")
-    return datetime.fromtimestamp(json.loads(base64.b64decode(payload_base64))["exp"])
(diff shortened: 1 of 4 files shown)


LS80:
Closing in favour of a new PR targeting the nexus branch — this update now relies on InfoTagVideo setter methods that were added in Kodi Nexus (v20), so it's no longer compatible with Matrix.

PR #4832, opened 2026-09-20
[plugin.video.sermonindex] 1.0.0

sermonindex-dev:
Same add-on as the Omega submission (#4831), for Piers.


basrieter:
If it is in the Omega branch, it will automatically be available in the Piers branch. So if there are no code differences between the 2, please only submit a PR to the Omega branch.

